PR #24, opened 2020-09-13
Lua rewrite

lukas-reineke:
Complete rewrite in lua

* does not need embedded nvim instance
* does not need to read the file from disk
* much faster
* can run on whole file on every change

But it doesn't use `indentexpr` anymore. The indent level of the next non blank line is used.
In 99% of cases this is good enough. And the speed improvements we get from not needing to call into vimscript land is massive.


plugin/indent_blankline.vim
```diff
@@ -29,8 +28,6 @@ command! IndentBlanklineToggleAll call indent_blankline#commands#ToggleAll()
 
 augroup IndentBlanklineAutogroup
     autocmd!
-    autocmd Syntax * IndentBlanklineRefresh
-    autocmd BufWritePost * IndentBlanklineRefresh
-    autocmd FileChangedShellPost * IndentBlanklineRefresh
     autocmd OptionSet shiftwidth,tabstop IndentBlanklineRefresh
+    autocmd FileChangedShellPost,Syntax,TextChanged,TextChangedI,WinScrolled * IndentBlanklineRefresh
```
romgrk, line 32: You need a guard for `WinScrolled` unless your requirements are *nighlty build only*.

romgrk:
Do you plan to merge soon?

plugin/indent_blankline.vim
lukas-reineke: I am waiting to merge until 0.5 is released

doc/indent_blankline.txt
```diff
@@ -73,6 +73,71 @@ g:indent_blankline_space_char                  *g:indent_blankline_space_char*
...
+g:indent_blankline_sapce_char_highlight*g:indent_blankline_sapce_char_highlight*
+
+    Specifies the space character highlight.
+    Not used if |g:indent_blankline_space_char_highlight_list| is not empty.
+
+    Default: 'Whitespace'                                                    ~
+
+    Example: >
+
+        let g:indent_blankline_sapce_char_highlight = 'Error'
```
Sh3Rm4n, line 113:
```suggestion
        let g:indent_blankline_space_char_highlight = 'Error'
```

:)

lukas-reineke: thanks

adelarsq:
Amazing job @lukas-reineke 🎉👏🏻

Issue #52, opened 2021-03-03
Only works for files in current directory


EtiamNullam:
This plugin seems to only work on files in current directory. Looks like a problem with slashes or line ending characters on Windows.

Does not work for any of these paths:
```
.\Example.cs
Desktop\Example.cs
D:\Example.cs
..\Example.cs
~\Example.cs
```

Works fine only when file is opened directly like `nvim Example.cs`. Executing `:cd ..` after it is enough to make it stop working again.

OS: Windows 10
Neovim: v0.4.4

Minimal vimrc:
```viml
call plug#begin('~/.vim-plug')

    Plug 'lukas-reineke/indent-blankline.nvim'
    Plug 'Yggdroot/indentLine'
    Plug 'sheerun/vim-polyglot'

call plug#end()
```

## Some more info that might be useful according to troubleshooting in other issues

- Result of `verbose setlocal indentexpr?` (in typescript file):

```viml
indentexpr=GetTypescriptGraphQLIndent()
	Last set from ~\.vim-plug\vim-polyglot\after\indent\typescript.vim line 37
```

- This prints "hello" as expected:
```viml
:echon rpcrequest(g:indent_blankline_nvim_instance, 'nvim_eval', '"hello"')
```

- Nothing in `:messages` after running with this line in vimrc:
```viml
let g:indent_blankline_debug = v:true
```
...unless when its a VimL file:
```
indent-blankline encountered an error while finding matches: Vim:E492: Not an editor command: viminit.vim
```
...or when you open some help page (in this case simple `:help`):
```
indent-blankline encountered an error on refresh: Vim(call):E12: Command not allowed from exrc/vimrc in current dir or tag search
indent-blankline encountered an error while finding matches: Vim:E492: Not an editor command: eovimNeovimshare^@vim^Muntimedochelp.txt
```

lukas-reineke:
Sorry I can't support windows for version 1.
I don't have access to a windows install, so I can't test anything.

If you update to Neovim 0.5 you can use the lua branch. That should work fine on windows. (also no need for `Yggdroot/indentLine` anymore, plus a lot of other improvements)


As I am only waiting for Neovim 0.5 to become stable before the lua rewrite will be the default, I am not really interested in big changes to the current version.
But if you can fix it yourself, feel free to open a PR.

EtiamNullam:
That's sad but it was to be expected.

[...]

PR #53, opened 2021-03-06
Fix issue #52 where plugin will not work for files that are not in current working directory

EtiamNullam:
Works for me so far. Will preserve user `shellslash` setting. Thanks for the help.

See issue #52 for more info.


EtiamNullam:
Rebased to `develop` as I'm targeting it.

lukas-reineke (approved):
LGTM, thank you

PR #68, opened 2021-03-23
Context highlight list

rhino1998:
This PR adds a list for defining different highlights at different levels with context highlighting. Currently there is no synchronization between brace/paren/other scope delimiter and the indent line color. It is just coincidental that the colors of the braces and the line match in the second image.
image
image

### DIFF
--- a/doc/indent_blankline.txt
+++ b/doc/indent_blankline.txt
@@ -367,7 +367,7 @@ g:indent_blankline_context_highlight_list*g:indent_blankline_context_highlight_l
 
     Example: >
 
-        let g:indent_blankline_context_highlight = 'Error'
+        let g:indent_blankline_context_highlight_list = ['Error', 'Warning']
 
 ------------------------------------------------------------------------------
 g:indent_blankline_context_patterns      *g:indent_blankline_context_patterns*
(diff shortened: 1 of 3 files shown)


doc/indent_blankline.txt
```diff
@@ -354,6 +354,21 @@ g:indent_blankline_context_highlight    *g:indent_blankline_context_highlight*
...
+    each indentation level.
+    Ignored if the value is an empty list.
+
+    Only used when |g:indent_blankline_show_current_context| is active
+
+    Default: '[]'                                                         ~
+
+    Example: >
+
+        let g:indent_blankline_context_highlight = 'Error'
```
lukas-reineke, line 370: example is wrong

lukas-reineke (changes requested):
Looks good, Thank you.

You are missing one instance of `context_highlight` for trailing indent on line 112
You'll need to calculate the index with `math.ceil(#virtual_text / 2) + 1)`

doc/indent_blankline.txt
rhino1998: oh, woops, fixed

lukas-reineke (approved):
Thank you

Issue #74, opened 2021-03-26
Lua: Allow display of listchars other than eol


timbedard:
indentLine would leave listchars like `tab` displayed, rather than replace them with its lines as indent-blankline is doing right now. With the `tab` listchar set, I think showing it is the expected behavior.

Similarly, when a line contains only spaces and `trail` is set, I would expect the list character to be displayed instead of the indent lines.

lukas-reineke:
both sound like good options 👍 
I'll look into it

WhyNotHugo:
I've managed to bisect the issue, and the regression start in d9f682939ac8a59b1784802b0c69e82177a4972a.

However, I'm not proficient enough in lua to further pinpoint what's up there.

lukas-reineke:
It's not really a regression. The whole functionality of the plugin changed.
Before using the new overlay virtual text, the plugin only did empty lines. And by definition, those didn't have trailing spaces.

lcrockett:
Personally i'd prefer to keep being able to see the difference between tabs and spaces using the `listchars` defined characters.

Would it be possible to accomplish this in any way ?

ek234:
hey,
I am a comp sci student. I wanted to start open-source contributions.

Can I work on this issue/feature request?

lukas-reineke:
Hey @ek234,
sure! Always grateful for help.
What I am thinking for this feature is an option that is a list of all listchars that should take priority over indent-blankline virtual text.
So for example `let g:indent_blankline_foo = ["eol", "tab"]` would not overwrite tab characters and end-of-line characters.

I think we need to care about `eol`, `tab`, `space`, `trail` and `lead`, but I haven't looked into it too much.

It will get complicated when there are cases where a line has a mix of showing indentation lines and not showing them. Like mixed indentation with spaces and tabs. It might make sense to just not support this for now, not sure.

If you start working on this, open a draft PR early. I'll review the code and help you, so you don't spend a lot of time on something I won't want to merge.

ek234:
Hi @lukas-reineke,

Even I thought of implementing it with a similar approach. However, an issue came up.

Consider in listchar that tab is set to anything other than '|\ ', for example, let it be "tab:>-" [like in the nvim documentation].
Then all the non blank lines (which actually contain tabs for indenting) will use this scheme and the plugin wouldn't write over it. But, the blank lines will still use the regular hard coded scheme.

>--->---this is what it would look like
|     |     |
[...]

PR #114, opened 2021-05-28
Optional skipping of space char if the whitespace char is not there

qualious:
**New option:**

`g.indent_blankline_skip_space_char_on_tab = <true/false>`

**Motivation:**

Given `tabstop=4` and the inserted character is a  tab, I want to visually distinguish my indentation indicator from `4 spaces`.

This is overall an important feature to have. For example: I'm working on an old python project at the moment which doesn't have any enforced space/tab standard. So, you know, the indentation get's a bit fucky sometimes. And we don't want to destroy the git history by formatting the whole thing. 

This is here for me to visually see it when that happens. I'm sure there are lots of other benefits that can be thought of. I feel like this is closer to intended implementation of such future.

**Screenshots:**

If indentation is tabs:

tabs

If Indentation is spaces:

spaces

Feedback is welcome! I'm not really knowledgeable about the rest of the code so I tried to do the modifications as silently as I can. Feel free to modify the PR if the LOC can be lowered.

### DIFF
--- a/lua/indent_blankline/init.lua
+++ b/lua/indent_blankline/init.lua
@@ -54,6 +54,7 @@ local refresh = function()
     local end_of_line = vim.g.indent_blankline_show_end_of_line
     local end_of_line_char = vim.fn["indent_blankline#helper#GetListChar"]("eol", "")
     local strict_tabs = vim.g.indent_blankline_strict_tabs
+    local skip_space_char_on_tab = vim.g.indent_blankline_skip_space_char_on_tab
 
     local tabs = vim.bo.shiftwidth == 0 or not expandtab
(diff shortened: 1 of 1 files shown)


lukas-reineke:
Thank you for the PR, I appreciate it.

But this is not the solution I want to go with. There is an issue for this #74 that covers all list chars (tab is a list char as well).
I want to add an option with a list of list chars, that defines what takes priority over the indent characters.
So it displays the actual character you set.
Also, just taking the first character to decide if its space or tab indentation is not good enough, the plugin supports mixed indentation.

Sorry, but I will close the PR. I hope I have some time soon to work in this.

PR #116, opened 2021-06-02
draft | lua: added support for customising indent

ek234:
This is a draft PR.
related to issue#74
With this, the user will be able to customize the appearance of the indent in a way similar to the tab in listchar in nvim.

new variables:
indent_blankline_char_first : (default is '|')
indent_blankline_char_middle : (default is ' ')
indent_blankline_char_end : (default is '')

These define the PATTERN the indent follows.

if char_end is empty then
> always put the char_first once
> put char_middle as many times to fill the tab length

if char_end in non empty then
> put char_first if more than one space is left in the tab length
> repeatedly put char_middle until only one space is left
> always put char_end

Note: the values of ' ' and '|' are special place holders. if ' ' is used then indent_blankline_char_space is inserted. and if '|' is used, then indent_blankline_char or indent_blankline_char_list is referred.
For any other char, the char itself is used.**

In the near future, the plan is to remove the dependency on these variables and use listchar itself. (thank you @lukas-reineke XD)

** I have implemented it in this way is because the user may have a char_list defined. Here, if the user uses '|', they would like that to represent a char from char_list. So I kept '|' as a placeholder for either the corresponding char from indent_blankline_char_list (if defined) or the default indent_blankline_char. 

The user may want to display ' ' itself. In this case, we can have '\ ' correspond to space[not implemented yet; but it should be easy to add]. Similarly '\|' can work with '|'.

### DIFF
--- a/lua/indent_blankline/init.lua
+++ b/lua/indent_blankline/init.lua
@@ -139,7 +139,7 @@ local refresh = function()
 
     local get_virtual_text = function(indent, extra, blankline, context_active, context_indent)
         local virtual_text = {}
-        for i = utils.if(first_indent, 1, 2), math.min(math.max(indent, 0), max_indent_level) do
+        for i = utils._if(first_indent, 1, 2), math.min(math.max(indent, 0), max_indent_level) do
             local context = context_active and context_indent == i
             -- first char in indent:
                 if( i == 1 and blankline and end_of_line and #end_of_line_char > 0 ) then
--- a/lua/indent_blankline/init.lua
+++ b/lua/indent_blankline/init.lua
@@ -67,58 +67,176 @@ local refresh = function()
 
     local get_virtual_text = function(indent, extra, blankline, context_active, context_indent)
         local virtual_text = {}
-        for i = 1, math.min(math.max(indent, 0), max_indent_level) do
-            local space_count = space
+		-- first indent:
+
+        for i = 2, math.min(math.max(indent, 0), max_indent_level) do
             local context = context_active and context_indent == i
-            if i ~= 1 or first_indent then
-                space_count = space_count - 1
-                table.insert(
-                    virtual_text,
-                    {
-                        utils._if(
-                            i == 1 and blankline and end_of_line and #end_of_line_char > 0,
-                            end_of_line_char,
-                            utils._if(
-                                #char_list > 0,
-                                utils.get_from_list(char_list, i - utils._if(not first_indent, 1, 0)),
-                                char
-                            )
-                        ),
-                        utils._if(
-                            context,
-                            utils._if(
-                                #context_highlight_list > 0,
-                                utils.get_from_list(context_highlight_list, i),
-                                context_highlight
-                            ),
-                            utils._if(
-                                #char_highlight_list > 0,
-                                utils.get_from_list(char_highlight_list, i),
-                                char_highlight
-                            )
-                        )
-                    }
-                )
-            end
-            table.insert(
-                virtual_text,
-                {
-                    utils._if(blankline, space_char_blankline, space_char):rep(space_count),
-                    utils._if(
-                        blankline,
-                        utils._if(
-                            #space_char_blankline_highlight_list > 0,
-                            utils.get_from_list(space_char_blankline_highlight_list, i),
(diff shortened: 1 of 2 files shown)


lukas-reineke:
I think you are going in the wrong direction with this.
This is not what issue #74 is about.
I don't want a new way to customize the appearance of the indentation.
The point of the ticket is to be able to see listchars instead of the indentation.

For the tab listchar, this does require some way in how the indentation is generated, but your approach is not the same as the listchar logic.
Please read `:h listchars`, especially `:h lcs-tab`

```
							*lcs-tab*
	  tab:xy[z]	Two or three characters to be used to show a tab.
			The third character is optional.

	  tab:xy	The 'x' is always used, then 'y' as many times as will
			fit.  Thus "tab:>-" displays:
				>
				>-
				>--
				etc.

	  tab:xyz	The 'z' is always used, then 'x' is prepended, and
			then 'y' is used as many times as will fit.  Thus
			"tab:<->" displays:
				>
				<>
				<->
				<-->
				etc.

			When "tab:" is omitted, a tab is shown as ^I.
```

ek234:
Thanks for the review.

I misunderstood the purpose of #74. I thought @timbedard  wanted to customize the indent according to their tab listchar ( maybe I read it like this because this was something I wanted to have ).

However, what I have implemented till now is a general total customization of the indent.
AND if we want to print different listchars differently, we need some way of printing it differently.
Now, I just need to read the underlying char, and print it accordingly [which should be much easier with this customized indent]


I am sorry if my wordings were confusing, but I did refer to `:h lcs-tab`. char_first, char_middle, char_end are the three characters of `lcs-tab` [third char is optional]. I have used three different variables for now as it is a bit more convenient to me. I will surely remove them before finalizing the PR.
When we encounter a tab character, we can have this set to the tab listchar pattern itself. And so on for other special chars.

[ Additionally, we can change this to any custom pattern too. This is helpful when the user wants to set the indentation according to their preferences but not the list char. ]

Let me know if you find customization a helpful feature in general. As I said, this is a feature I wanted to have in the first place. I believe others might find it useful too.

lukas-reineke:
> if we want to print different listchars differently, we need some way of printing it differently.
Now, I just need to read the underlying char, and print it accordingly [which should be much easier with this customized indent]

Yes I agree. Using the listchar tab system to create the indent for everything could work, and make it easier to support listchars.
But it is a much bigger task. There are a lot of edgecases you will need to cover if you do this.

> I am sorry if my wordings were confusing, but I did refer to :h lcs-tab

:+1: yeah it was written a bit confusing, but I get it now. It is the same

ek234:
@lukas-reineke, should I continue with this approach? If there is something else you would prefer done differently, then please do let me know.

Thanks

lukas-reineke:
> should I continue with this approach? If there is something else you would prefer done differently, then please do let me know.

It's up to you. It will be difficult to get this to work correctly 100%. And I am not going to merge it until it does.
But if you think you can do it, go for it.

Also keep in mind that we can't remove features that are already there, like `show_trailing_blankline_indent` and `show_first_indent_level`

ek234:
okay i'll try to explore this a bit more.

thanks

ek234:
I just had a small doubt.
Consider space indent with shiftwidth = 4. in listchar, tab is `<.>`, ie, a complete tab is represented as `<..>`

If we have a line with 2 spaces followed by a tab followed by two spaces, currently we see:
> |␣␣␣|␣␣␣
As the tab is completely hidden.

In this case, is the expected behavior this:
> |␣<..>␣␣
that is, do I simply write the tab list char on top of the indents?

lukas-reineke:
yeah, this is where it gets complicated. I'd say, display `|␣␣␣|␣␣␣` as the default, but when the user wants to display the tab listchar, it should be `|␣<..>␣␣`

As you noted in #120 irregular indentation with mixed tabs and spaces currently breaks. It would be nice to handle that as well with the same logic.
I just didn't bother until now because it is very rare and hard to fix.

ek234:
yeah, I agree this is quite a rare case. However, this case is important for completely realizing #74.

I have added some approaches to #120. I think we will have to fix it separately. Also, it will have to be resolved before we can proceed with #74.

Issue #120, opened 2021-06-08
indents with both space and tabs hide text


ek234:
Consider a nonblank line containing a space followed by a tab followed by some text.

Here, utils.find_indent() returns `1, true`. that is, 1 indent and extra=true.

However, a space followed by a tab takes the same space as a tab. So, text continues after the tab. But, due to extra=true, an indent is drawn right after the tab.

Due to this, the first letter of the text is hidden. [this was a simple case; more text might be hidden in other cases]

strict tab also causes some issues.

ek234:
There are a couple of ideas to fix this:

1. In utils.find_indent() iterate over the whitespace_string char by char. Keep a counter of space characters which is reset to 0 whenever a tab is encountered.
This is easier to understand while reading the code but doesn't look that good.

2. Otherwise, we can add variables:
> local _, 1spacetab = whitespace_string:gsub("<space><tab>", "")
> local _, 2spacetab = whitespace_string:gsub("<space><space><tab>", "")
> local _, 3spacetab = whitespace_string:gsub("<space><space><space><tab>", "")

And calculate the correct indent with these extra variables.
I, personally, would prefer this.

Let me know what you would prefer. If you want, I might be able to implement this by today or tomorrow.

lukas-reineke:
2 will break again if you have more than one tab in the indentation, like `|␣<..>␣␣|␣<..>␣␣`

It has to be option 1.

> Keep a counter of space characters which is reset to 0 whenever a tab is encountered.

You still need to count spaces as well. Just subtract the count until the last tabstop when you encounter a tab char.
And `strict_tabs` should use the tab count until it hits the first space char.

lukas-reineke:
Properly fixed now with #121

PR #121, opened 2021-06-09
Fixed irregular indentation

ek234:
fixed find_indent() according to #120

### DIFF
--- a/lua/indent_blankline/init.lua
+++ b/lua/indent_blankline/init.lua
@@ -121,7 +121,7 @@ local refresh = function()
             )
         end
 
-        if ((blankline and trail_indent) or extra) and (first_indent or #virtual_text > 0) then
+        if ((blankline or extra) and trail_indent) and (first_indent or #virtual_text > 0) then
             local index = math.ceil(#virtual_text / 2) + 1
             table.insert(
                 virtual_text,
--- a/lua/indent_blankline/utils.lua
+++ b/lua/indent_blankline/utils.lua
@@ -117,17 +117,25 @@ M._if = function(bool, a, b)
 end
 
 M.find_indent = function(line, shiftwidth, strict_tabs)
-    local _, whitespace_count = line:find("^%s+")
-    if not whitespace_count or whitespace_count == 0 then
-        return 0, false
+    local indent = 0
+    local spaces = 0
+    for ch in line:gmatch(".") do
+        if ch == "	" then
+            if strict_tabs and indent == 0 and spaces ~= 0 then
+                return 0, false
+            end
+            indent = indent + math.floor(spaces / shiftwidth) + 1
+            spaces = 0
+        elseif ch == " " then
+            if strict_tabs and indent ~= 0 then
+                return indent, true
+            end
+            spaces = spaces + 1
+        else
+            break
+        end
     end
-    local whitespace_string = line:sub(1, whitespace_count)
-    local _, spaces = whitespace_string:gsub(" ", "")
-    local _, tabs = whitespace_string:gsub("	", "")
-    if strict_tabs and tabs > 0 then
-        return tabs, spaces > 0
-    end
-    local indent = tabs + (spaces / shiftwidth)
+    indent = indent + math.floor(spaces / shiftwidth)
     return indent, spaces % shiftwidth ~= 0
 end
 


lukas-reineke:
This breaks indentation with just spaces

ek234:
Can you please give me a line that breaks the indentation? It seems to work for me.

lukas-reineke:
any line indented with just spaces, when you turn on `strict_tabs`

ek234:
when `strict_tabs` is set, then only tabs are used for detecting indent right?
So, if the line contains no tabs then the expected behavior should be to not indent anything right?

lukas-reineke:
> when strict_tabs is set, then only tabs are used for detecting indent right?

no this is wrong. Here is the documentation

```
    When on, if there is a single tab in a line, only tabs are used to
    calculate the indentation level.
    When off, both spaces and tabs are used to calculate the indentation
    level.
    Only makes a difference if a line has a mix of tabs and spaces for
    indentation.
```

ek234:
ahh, I see.. I'll push the corrected code in some time

lukas-reineke:
that looks good :+1: 

can you do a couple small things.
1. rename `new_spaces` to just `spaces`
2. format the file with [luafmt](https://github.com/CurtisFenner/luafmt)
3. squash all your commits into one commit

ek234:
First, I added the condition to correct the implementation of `strict_tabs` to the previous code.
However, this made that specific code a bit bulky and unappealing.
So I re-implemented the logic like this:
```
M.find_indent = function(line, shiftwidth, strict_tabs)
    local indent = 0
    local new_spaces = 0
    for ch in line:gmatch(".") do
        if ch == "	" then
            if strict_tabs and indent == 0 and new_spaces ~= 0 then
                return 0, false
	    end
            indent = indent + math.floor( new_spaces / shiftwidth ) + 1
            new_spaces = 0
        elseif ch == " " and ( not strict_tabs or indent == 0 ) then
            new_spaces = new_spaces + 1
        else
            break
        end
    end
    indent = indent + math.floor(new_spaces / shiftwidth)
    return indent, new_spaces % shiftwidth ~= 0
end
```
`new_spaces` is a counter of spaces since the last tab character. `indent` stores the definite number of indents read till now.
Keeping the number of tabs and space indents separately wasn't making any difference in this implementation so I decided to store them together in `indent` to clean the code and reduce calculations.

Implementing of strict_tabs is done in two parts:
1. if the line starts with some spaces and then has at least one tab:
>> if strict_tabs and indent == 0 and new_spaces ~= 0 then
>>     return 0, false
>> end
2. if the line starns with some tabs and then has at least one space
>>  elseif ch == " " and ( not strict_tabs or indent == 0 ) then

Please let me know if there should be any other modifications.

ek234:
okay i will do that.

btw can ya please tell me how to squash the commits to one XD.. I am super new to git.

thanks

ek234:
Ohh, that's super handy. Thanks.

I'll reformat and push.

ek234:
>     2. format the file with [luafmt](https://github.com/CurtisFenner/luafmt)

I did everything you mentioned. One issue I faced was that luafmt converted all the indents in the file to tabs. The rest of the codebase was space indented so I substituted the indent to spaces.
Let me know if you want to keep that as tabs.

lukas-reineke:
sorry i linked the wrong luafmt :man_facepalming: 
i didn't even know there were 2

here is the correct link https://github.com/trixnz/lua-fmt

ek234:
done

lukas-reineke:
I found another small problem

with `strict_tabs` on, a line with first tab indentation, and then space indentation, needs to return `extra = true` (the second value that's returned)

eg. this line `		          fooo` (2 tabs, then spaces)
should look like this

Screen Capture_select-area_20210611142300

but in your PR it looks like this

Screen Capture_select-area_20210611142323

ek234:
I did consider this case in my testing. Here, strict_tabs is on and both spaces and tabs are present. So, according to the docs, only tabs will be considered in the indent.
Therefore, the space which comes after the tabs isn't even a part of the indent. and so, I return extra=false.

I understand that `extra==true` means that there is some more indentation in the line which hasn't been counted in the `indent` variable.
But if a user sets strict_tabs then they are not expecting spaces to be a part of indent in the first place right? Else it will imply a situation where 3tabs is indented the same as 2tabs+4spaces. I felt that a user who set strict tabs would not want this.

Please let me know what you think.

Ps. what color scheme are you using? I really like that shade of green:P

lukas-reineke:
> But if a user sets strict_tabs then they are not expecting spaces to be a part of indent in the first place right? Else it will imply a situation where 3tabs is indented the same as 2tabs+4spaces. I felt that a user who set strict tabs would not want this.

Yes, you are right, kindof.
It should depend on `g:indent_blankline_show_trailing_blankline_indent`

It basically is a trailing indent. It should match what the other lines are doing.

Take for example this code.

```c
#include <stdio.h>

int main(int argc, char *argv[])
{
	printf("test string %s:%s():%d\n",
	       __FILE__, __func__, __LINE__);

	printf("hello");
}
```

in your PR it would look like this with `show_trailing_blankline_indent` on

Screen Capture_select-area_20210611160638

But with it should look like this

Screen Capture_select-area_20210611160717

But with `show_trailing_blankline_indent` off, it makes sense that it looks like you have it now

Screen Capture_select-area_20210611161207

lukas-reineke:
The color scheme is my version of one dark
https://github.com/lukas-reineke/onedark.nvim

ek234:
Oh, I see...

So should I replace
```
[current]          if ((blankline and trail_indent) or extra) and (first_indent or #virtual_text > 0) then
```
with
```
[suggestion]       if ((blankline or extra) and trail_indent) and (first_indent or #virtual_text > 0) then
```
[in file init.lua]

Or is there any case where even though trail_indent is false, extra causes another indent character to be printed?

lukas-reineke:
> So should I replace...

Yes :+1: 

This looks good now.
I don't have that much time today. I'll test a bit more tomorrow, but I think this covers everything now.
Good job

ek234:
aight I squashed everything

thanks

ek234:
Is everything alright?

lukas-reineke:
Yeah sorry just didn't have time
all good
thank you

ek234:
Thanks!!!
This was my first proper PR, and it was really fun working with you.

PR #134, opened 2021-07-01
Update README.md

ofsaleem:
indent-blankline yelled at me for having IndentLine installed, should probably remove the lines in the Readme that mention its recommended to have both installed, since it doesn't seem to be the case anymore.

### DIFF
--- a/README.md
+++ b/README.md
@@ -45,9 +45,6 @@ endif
 
 ## IndentLine dependency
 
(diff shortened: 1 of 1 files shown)


ofsaleem:
nevermind, i was on the wrong branch, seems like youve already cleared it up for the lua branch, ty

PR #173, opened 2021-08-18
Update indent_blankline.txt

JuanVqz:
I wanted to exclude ruby files but didn't work as an array.

besides that, how can I change the `char` color? I didn't understand the help file.

Thank you for your plugin!

### DIFF
--- a/doc/indent_blankline.txt
+++ b/doc/indent_blankline.txt
@@ -328,7 +328,7 @@ g:indent_blankline_filetype_exclude      *g:indent_blankline_filetype_exclude*
 
     Example: >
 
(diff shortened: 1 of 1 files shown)


lukas-reineke:
This change is wrong.
The example is in Vimscript, the array syntax is `[]`

You need to change the `IndentBlanklineChar` highlight group to change the character color. Take a look at `:help :highlight`.

JuanVqz:
you are right! 

I used this one.
```vim
lua require('indent_blankline').setup {
  char = '┊',
  filetype_exclude = {'help', 'ruby'}
}
```

PR #187, opened 2021-08-29
Allow for display of listchars in indentation overlay.

Daxtorim:
My attempt at implementing #74. I switched from generating the virtual text by just adding spaces until it has the appropriate length to an approach using the actual leading whitespace content of each line.

Currently I am replacing each tab (`^I`) with a `ta..ab` string of length of the width of the tab. Every other whitespace character gets replaced with `s`. Here is an intentionally messy file without this plugin, just my defined `listchars`:
`set list listchars=tab:›\ ,trail:~,space:⋅,nbsp:⍽,extends:>,precedes:<`

image

And here it is with my PR:
image

I still need to source the actual `listchars` of the user and replace the placeholders of course, but this is just a draft to let you know someone is working on it.

UPDATE: (Different font to avoid ligatures; that's gonna be a headache to fix, if even possible)
I changed `listchars` to be `tab:>-<` and added `eol:↴` for the purpose of this screenshot.
image
edit: welp, maybe displaying isn't quite ready yet ^^ (FIXED)

I can't reliably source the user's `listchars` for tab characters because Lua 5.1 doesn't have proper UTF-8 support. Everything works on bytes, which breaks down when you need to extract a single character from a string with multi-byte long characters (see the comment in the code).
I added three new variables for that:
```
vim.g.indent_blankline_tab_char_start
vim.g.indent_blankline_tab_char_fill
vim.g.indent_blankline_tab_char_end
```
They are each of the three characters you would put in your `listchars` as well. 

Oh, and I broke Treesitter support for the indents, since it wasn't compatible with my approach. I'll see if I can add that later again.

TODO:

- [X] Get actual line content as template
- [X] Replace dynamic-width whitespace with fixed-width placeholders
- [X] Display indentation char over placeholder
- [X] Get blank lines to behave
- [X] Replace placeholders with user defined characters
- [x] Display `trail` characters when line contains *only* whitespace
- [x] Clean up the files' whitespace and remove now obsolete stuff
- [ ] Add documentation for new variables
- [ ] Fix Treesitter support (Maybe? I never used that anyway)

### DIFF
--- a/lua/indent_blankline/init.lua
+++ b/lua/indent_blankline/init.lua
@@ -47,27 +47,27 @@ M.setup = function(options)
         vim.g.indent_blankline_space_char_highlight_list
     )
 
-	-- Neovim still uses Lua 5.1 which doesn't have UTF-8 support (was only introduced in Lua 5.3)
-	-- Because of that there is no reliable way to differentiate between tab listchars
-	-- string.sub() acts on bytes, not "characters"
-	-- consider `listchars=tab:›\ ‹`; these chars are SINGLE RIGHT-POINTING ANGLE QUOTATION MARK and left equivalent
-	-- Each is a 3 byte long character, the entire sequence is <e2><80><b8><20><e2><80><b9>
-	-- string.sub() will grab each byte individually instead of a whole character
-	-- without a proper UTF-8 library there is too much guess work to finding the correct sequence
-	-- Hence, we don't source listchars at all and rely on global plugin specific variables
+    -- Neovim still uses Lua 5.1 which doesn't have UTF-8 support (was only introduced in Lua 5.3)
+    -- Because of that there is no reliable way to differentiate between tab listchars
+    -- string.sub() acts on bytes, not "characters"
+    -- consider `listchars=tab:›\ ‹`; these chars are SINGLE RIGHT-POINTING ANGLE QUOTATION MARK and left equivalent
+    -- Each is a 3 byte long character, the entire sequence is <e2><80><b8><20><e2><80><b9>
+    -- string.sub() will grab each byte individually instead of a whole character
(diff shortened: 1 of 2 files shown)


lukas-reineke:
First, thanks for taking a look into this.
I get what you are going for, but you are touching way too many places, and because of that a lot of stuff breaks.
Backwards compatibility is important to me. Everything has to work like it did before.

Really you should only need to extend `find_indent` to also return the `virtual_string` you made (and you don't need placeholders, just use the correct characters directly).
Then when creating the virtual text, pop the strings from there. Everything else can stay as it is.

> Lua 5.1 doesn't have proper UTF-8 support

Lua doesn't, but Neovim does. You can do `:lua print(vim.inspect(vim.fn.split("› ‹", '\\zs')))`

Daxtorim:
Yes, I got a little carried away while experimenting.😅 I see what you are saying, and now that I have a working proof of concept it shouldn't be too difficult to take the relevant parts and put add them on top of master without ripping out half the original code.

>Lua doesn't, but Neovim does. You can do `:lua print(vim.inspect(vim.fn.split("› ‹", '\\zs')))`

Hmm, that will certainly let me source `listchars` easily. But if I were to not create a template and instead use tab `listchars` directly I would need to split the string into a list of characters, then grab the correct number from the list, then reconstruct the substring again. I might as well store the virtual string as a list directly. More experimentation ahead, yay.

lukas-reineke:
> I might as well store the virtual string as a list directly

Yeah, I think that is the best solution :+1:. The `get_virtual_text` function is async, so you can't use `vim.fn` inside, that would make working with a unicode string a pain again.
Return a table with the correct characters from `find_indent` and pop them off in `get_virtual_text`.
You can use `vim.list_slice()`

Daxtorim:
Superseded by #190

PR #190, opened 2021-08-31
Allow for display of listchars in indentation overlay. (refactor)

Daxtorim:
My attempt at implementing #74. Now fully backwards compatible.

`set list listchars=tab:<->,space:⋅,trail:~,eol:↴`
Ligatures are handled by the terminal, not this plugin.
image

The new behavior can be activated with `g:indent_blankline_show_whitespace`, which defaults to `false`. There won't be any change to existing behavior when updating without user intervention.
I also added several new variables (with documentation) to customize the appearance of the overlay text. They all default to their `listchars` values, or have appropriate fallbacks. 
```
g:indent_blankline_trail_char
g:indent_blankline_tab_char_start
g:indent_blankline_tab_char_fill
g:indent_blankline_tab_char_end
```

### DIFF
--- a/lua/indent_blankline/init.lua
+++ b/lua/indent_blankline/init.lua
@@ -28,18 +28,12 @@ M.setup = function(options)
         o(options.char_list, vim.g.indent_blankline_char_list, vim.g.indentLine_char_list, {})
     vim.g.indent_blankline_char_highlight_list =
         o(options.char_highlight_list, vim.g.indent_blankline_char_highlight_list, {})
-    vim.g.indent_blankline_space_char =
-        o(options.space_char, vim.g.indent_blankline_space_char, vim.opt.listchars:get().space, " ")
     vim.g.indent_blankline_space_char_highlight_list =
         o(options.space_char_highlight_list, vim.g.indent_blankline_space_char_highlight_list, {})
     vim.g.indent_blankline_space_char_blankline =
         o(options.space_char_blankline, vim.g.indent_blankline_space_char_blankline, vim.g.indent_blankline_space_char)
     vim.g.indent_blankline_space_char_blankline =
-        o(
-        options.space_char_blankline,
-        vim.g.indent_blankline_space_char_blankline,
-        vim.g.indent_blankline_space_char_highlight_list
-    )
+        o(options.space_char_blankline, vim.g.indent_blankline_space_char_blankline, " ")
     vim.g.indent_blankline_space_char_blankline_highlight_list =
         o(
         options.space_char_blankline_highlight_list,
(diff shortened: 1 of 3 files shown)


lukas-reineke:
Already looks much better!
There are a couple of things I'd like you to change.

1. we don't need `g:indent_blankline_trail_char`, `g:indent_blankline_tab_char_start`, `g:indent_blankline_tab_char_start` and `g:indent_blankline_tab_char_end`
   that is just duplicating functionality Neovim already has.
2. It makes it a bit more complex, but I think we should allow turning specify listchars on and off, instead of one global `g:indent_blankline_show_whitespace`
   it can be a table that maps to the listchar `{eol = true, trail = false, tab = true, ...}`
   with this, we could then also remove the `g:indent_blankline_show_end_of_line` option

Daxtorim:
For clarification:
1. Should I just remove the variables from `setup()` and instead source `listchars` within `refresh()` directly without creating global variables? Or what do you mean by "duplicating functionality Neovim already has"?
2. When a specific group is set to `false`, what should be displayed instead? `space_char`? An empty space? 
3. What do you want the defaults to be? Current default behavior would be this I think:
`show_whitespace = { space = true, tab = false, trail = false, eol = false }`
Do you want to keep that?

I swear I ran `luafmt` before committing, but I guess I didn't.

lukas-reineke:
> Should I just remove the variables from setup() and instead source listchars within refresh() directly without creating global variables?

yes :+1:
I think we should even remove `g:indent_blankline_space_char` and just use listchars for everything.

> When a specific group is set to false, what should be displayed instead? space_char?

Thinking about this more, maybe it would be better to just not allow to turn it off at all :thinking: 
Everyone can still change what is displayed from the list chars. What do you think?

> What do you want the defaults to be?

If we keep it, default should be on for all of them.

There are multiple programs called `luafmt` maybe you used another one
this is the one I use https://github.com/trixnz/lua-fmt

lukas-reineke:
> we could then also remove the g:indent_blankline_show_end_of_line option

Actually, I don't think we can. EOL is displayed instead of the indent char, I think that's the only listchar that should behave this way.

I think like it best if displaying listchars just replaces the current logic and can't be turned off.
We remove `g:indent_blankline_space_char`, that will be set from listchars as well.
No new options.

Daxtorim:
I just pushed new commits. I did remove `g:indent_blankline_show_end_of_line` and it works fine.

> We remove g:indent_blankline_space_char, that will be set from listchars as well.

I do think there is merit in keeping this one. It allows you to tell the plugin to display leading spaces of a line without showing a special character for **all** spaces globally.

Daxtorim:
> I think like it best if displaying listchars just replaces the current logic and can't be turned off.

In that case I would actually start to rip out old code for real now. `strict_tabs` and `strict_spaces` and the code in utils.lua handling those become irrelevant.

lukas-reineke:
> It allows you to tell the plugin to display leading spaces of a line without showing a special character for all spaces globally

That's a good point :+1: let's leave it.

> strict_tabs and the code in find_indent() handling it becomes irrelevant.

That's not correct. Strict tab also influences until where the indent characters are displayed.
This still needs to work.

```
strict tab
taabtaabtaabsssssssss
|aab|aab|aabsssssssss

no strict tab
taabtaabtaabsssssssss
|aab|aab|aab|sss|ssss


and with show trailing:

strict tab
taabtaabtaabsssssssss
|aab|aab|aab|ssssssss

no strict tab
taabtaabtaabsssssssss
|aab|aab|aab|sss|sss|
```

Daxtorim:
>>It allows you to tell the plugin to display leading spaces of a line without showing a special character for all spaces globally

>That's a good point 👍 let's leave it.

I take that back actually. Neovim has an extra `listchars` lead character, which is exactly for this purpose. I actually had that implemented in earlier commits and then removed it because regular Vim had me question my sanity when it suddenly threw an error when setting it. There is actually different info in `:help listchars` depending on whether you opened it in Vim or Neovim. Considering that this plugin doesn't work with regular Vim anyway, I think `g:indent_blankline_space_char` can go away after all and I can implement `listchars` lead again.

>Strict tab also influences until where the indent characters are displayed.

Ah, then I will leave it alone.

Daxtorim:
Ok, I think everything works now, `strict_tabs` included. 

I removed `show_whitespace` and all the conditionals with it. We now source `listchars` for everything if that value is set and otherwise fall back to the space value, which itself falls back to a regular space (" "). 
`g:indent_blankline_space_char_blankline` still exists for people who want a special blankline character that can't be set via `listchars`, but it defaults to a regular space now.

Say if you want me to squash the commits or if you take the history as it is.

lua/indent_blankline/init.lua
```diff
@@ -125,32 +117,44 @@ local refresh = function()
     local char_highlight_list = v("indent_blankline_char_highlight_list")
     local space_char_highlight_list = v("indent_blankline_space_char_highlight_list")
     local space_char_blankline_highlight_list = v("indent_blankline_space_char_blankline_highlight_list")
-    local space_char = v("indent_blankline_space_char")
     local space_char_blankline = v("indent_blankline_space_char_blankline")
+
+    local space_character = vim.opt.listchars:get().space or " "
+    -- source tab chars in a way that works with UTF-8 characters
+    local tab_characters = vim.fn.split(vim.opt.listchars:get().tab, "\\zs")
```
lukas-reineke, line 124: If there is no tab set, this resolves to `{ "n", "u", "l", "l" }` :smile: 
we should default to `{space_character}` here

lukas-reineke:
This is what I mean with oel
Screen Capture_select-area_20210901143805

There is no way to not display the starting eol characters, without turning all of them off.
I'd like to keep this, so just put the option back.

Everything else looks good :+1: awesome work

lukas-reineke:
> Say if you want me to squash the commits or if you take the history as it is.

squashing would be nice

lua/indent_blankline/init.lua
Daxtorim: The issue here is that without a tab character set in `listchars` this doesn't really work anyway since Vim will display a literal `^I` with a fixed width of 2 columns, regardless of your tabstop setting.

image
There is no handling for irregular indentation, so with `tabstop=4`, or anything over 2 really, the virtual text will be overlaid over actual line content.
image


The only scenario where this works out is when you set `tabstop=2`. I guess that is something that the user is probably aware of when they make the *active choice to remove default `listchars` without replacement*.  I wouldn't spent any energy trying to work out how to properly support this for other tabstop/shiftwidth settings.

Daxtorim:
Re-added the `show_end_of_line`, put a condition around the assignment of `tab_characters` and squashed the commits.

I'll leave it to you to update the README if you want to add a notice about the new behavior. 👍 ❤️

lukas-reineke (approved):
:tada: :tada: 

thank you a lot

timbedard:
This is amazing. @Daxtorim, thank you so much for tackling this.

ggustafsson:
Does this fix not include support for "trail"? I am unable to find any information on it.

Issue #59, opened 2021-03-18
`cursorline` highlight is not cleared after the cursor is moved


ChristianChiarulli:
I'm not really sure how to explain this but for some black spaces some of the highlight gets left behind

![image](https://user-images.githubusercontent.com/29136904/111677928-1b9e5d00-87f6-11eb-8110-fe5d83fc7aa2.png)

ChristianChiarulli:
And if I mouse click in the buffer it disappears.

lukas-reineke:
Is that cursorline highlight?
I can't reproduce this.

Rendering bugs are very probably bugs in Neovim. Can you try to reproduce this in a clean environment?

`nvim -u NORC`

and then run this, it should write `hello world` as virtual text in the first line.
```
:lua ns_id = vim.api.nvim_create_namespace("foo")
:lua vim.api.nvim_buf_set_extmark(0, ns_id, 0, 0, { virt_text={{'hello world', 'String'}}, virt_text_pos='overlay'})
```

Does it happen for that as well?

kristijanhusak:
I can confirm this happens to me also, randomly. Once I move around it clears itself. Maybe some redrawing is needed, not sure.

ChristianChiarulli:
Yea I’ll have to test without some plugins. I haven’t tried barebones yet. I’m glad someone else can confirm the issue.

timbedard:
Same issue.

@lukas-reineke I tried your virtual text experiment. It doesn't retain the highlight the same way that the blank lines do when using indent-blankline either with an empty config or with indent-blankline, whether the virtual text line has text or is blank.

lukas-reineke:
I'm still not sure what highlight we are talking about. Is it cursorline or visual selection?
Can you also share what terminal emulator / neovim version / OS / tmux etc you are using?

ChristianChiarulli:
It’s the cursorline highlight. No tmux, alacritty, arch Linux.

ahmedkhalf:
I am also having a similar issue when scrolling down (go file w/tabs).

![image](https://user-images.githubusercontent.com/36672196/111946107-4b01e200-8af4-11eb-8eb7-b89ed3d9a03d.png)

lukas-reineke:
:thinking: 
I still can't reproduce it at all. I tried with alacritty and kitty on arch. Different filetypes. Tabs and spaces. Different nvim versions.

[...]

PR #191, opened 2021-08-31
fix: clear hi of current line after moved to next

mitinarseny:
When scrolling down and up fast, it seems like it does not have time to refresh
image


mitinarseny:
Nevertheless, it does not solves the problem of remaining highlighting of the next line like in following example
image
Any ideas?

lukas-reineke:
This is a bug in Neovim

There is already a better workaround in the issue here #59 until Neovim fixes the underlying issue.

PR #201, opened 2021-09-02
Add option to automatically turn off if the 'list' option is off

zeertzjq:
This allows toggling two thing with one command, or one mapping (`yol` from `vim-unimpaired`), and reduces the need to keep a list of disabled filetypes because many filetypes (e.g. `help`) have `ftplugin`s that unset `list`.


lua/indent_blankline/utils.lua
```diff
@@ -68,6 +69,9 @@ M.is_indent_blankline_enabled =
         if g_enabled ~= true then
             return false
         end
+        if respect_list then
+          return vim.opt.list:get()
```
lukas-reineke, line 73: This should not overwrite the other checks.
if it is `false`, return `false`, if it is `true`, continue with the other checks.

And this function is memoized, you have to get the value in `init.lua` and pass it as an argument

plugin/indent_blankline.vim
```diff
@@ -31,6 +31,7 @@ lua require("indent_blankline").init()
 
 augroup IndentBlanklineAutogroup
     autocmd!
+    autocmd OptionSet list lua require("indent_blankline.utils").list_changed()
```
lukas-reineke, line 34: you don't need this `list_changed` function, just  add it to the other `OptionSet` autocommand. `IndentBlanklineRefresh` will do the right thing

lukas-reineke (changes requested):
Thanks, I like it
maybe it should even be on by default

zeertzjq:
> This allows toggling two thing with one command, or one mapping (`yol` from `vim-unimpaired`), and reduces the need to keep a list of disabled filetypes because many filetypes (e.g. `help`) have `ftplugin`s that unset `list`.

For users who don't have `'list'` set having this on by default changes the behavior.

lua/indent_blankline/init.lua
```diff
@@ -91,6 +92,8 @@ local refresh = function()
         not utils.is_indent_blankline_enabled(
             vim.b.indent_blankline_enabled,
             vim.g.indent_blankline_enabled,
+            vim.g.indent_blankline_respect_list,
```
lukas-reineke, line 95:
```suggestion
            v("indent_blankline_respect_list"),
```

lukas-reineke (changes requested):
right, `list` is not on by default. Then let's go with this off by default as well :+1: 

Code looks good now, just one last change

lukas-reineke (approved):
Thank you!

Issue #195, opened 2021-09-02
Weird indentation characters. How to disable them?


Pocco81:
I just updated the plugin and these thing are appearing:

![image](https://user-images.githubusercontent.com/58336662/131775850-2f7b1686-4ccd-4d72-abc5-7217f91cbdbe.png)

AFAIK this was introduced since/by this commit: bc35145af2e2a4874fc39e42c885c73d5cfc88e8. Everything was working fine before (a702f809ce85b16a83c44016bf10fe5a5765e98d):

![image](https://user-images.githubusercontent.com/58336662/131775950-2d48cc36-b6b9-4b1e-870a-e7c143057c6d.png)

Is this bug or is it "the new normal"? In case this was intended, is there a way to disable this?

lukas-reineke:
This is now the correct behavior.
It displays what it would display without indent-blankline installed.
To change this, you need to set `listchars`. Either `tab` or `space`, depending on what you are using. `:help lcs-tab` `:help lcs-space` (or `:help lcs-space`)

Pocco81:
> It displays what it would display without indent-blankline installed.

Why? Shouldn't overriding the `listchars` be the expected behavior? If I want to see the fillchars then I'd uninstall indent-blankline and vice-versa.

Daxtorim:
Because without showing the characters defined in `listchars` there is no way to tell what the leading whitespace looks like. If the original file uses spaces and now you come in and use tabs, you leave behind an absolute mess without even knowing about it.

As you can also see in your screenshot above, this plugin *will* override characters at the indentation level with your defined character. The reason why it isn't doing it at the first level is becuase you set `vim.g.indent_blankline_show_first_indent_level = false`. This value would default to `true` otherwise.

aperum:
I was also surprised by the new behavior. I'd expect listchars showing only if list is set.
I have a key to toggle list/nolist if i want to check my whitespaces. The current behavior of showing listchars even when list is not set seems counter intuitive to me.

Daxtorim:
> I'd expect listchars showing only if list is set.

Truel, I guess. I have `set list` in my .vimrc and basically never turn it off. I didn't really consider how other people interact with this option.

I addressed this in #203.

PR #203, opened 2021-09-02
Allowing to remove all listchars but keep indentation guides with 'set list/nolist'.

Daxtorim:
Addressing #195.

I renamed the option `g:indent_blankline_respect_list` to the more descriptive name `g:indent_blankline_disable_with_list`. 
1. If this option is set to true, setting `nolist` will essentially disable the plugin as was implemented in #201.
2. If this option is set to false (default), setting `nolist` will replace all `listchars` with empty spaces (like `nolist` without this plugin) but keep indentation guides.

### DIFF
--- a/lua/indent_blankline/init.lua
+++ b/lua/indent_blankline/init.lua
@@ -42,7 +42,8 @@ M.setup = function(options)
     )
     vim.g.indent_blankline_indent_level = o(options.indent_level, vim.g.indent_blankline_indent_level, 20)
     vim.g.indent_blankline_enabled = o(options.enabled, vim.g.indent_blankline_enabled, true)
-    vim.g.indent_blankline_respect_list = o(options.respect_list, vim.g.indent_blankline_respect_list, false)
+    vim.g.indent_blankline_disable_with_nolist =
+        o(options.disable_with_nolist, vim.g.indent_blankline_disable_with_nolist, false)
     vim.g.indent_blankline_filetype =
         o(options.filetype, vim.g.indent_blankline_filetype, vim.g.indentLine_fileType, {})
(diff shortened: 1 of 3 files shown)


lukas-reineke:
Beat me to it, I wanted to do the same :+1: 

I think if we rename the option, it should be `g:indent_blankline_disable_with_nolist`
code looks good

lua/indent_blankline/init.lua
```diff
@@ -124,23 +125,39 @@ local refresh = function()
...
-        tab_characters = {}
+        -- nolist is set, replace all listchars with empty space
+        list_chars = {
+            space_char = " ",
+            trail_char = " ",
+            lead_char = " ",
+            tab_char_start = " ",
+            tab_char_fill = " ",
+            tab_char_end = " ",
+            eol_char = char
```
lukas-reineke, line 158: actually this might be confusing, even if it does the same thing.
Just set it to `""`

Daxtorim: Haha, I forgot already that there is a check to see of that char exists.
`""` won't work however, because Lua interprets even a zero-length string as `true`. I will set it explicitly to `nil`so it doesn't look like it was forgotten.

Same with `tab_char_end`, setting it to `nil` will result in the shorter branch inside `find_indent()`. That optimization will speed up the code by at least 10x 😎

Daxtorim:
Well, one should definitely commit before force-pushing 🙃
Addressed your suggestions, though.

lua/indent_blankline/init.lua
lukas-reineke: :rocket:

lukas-reineke (approved):
Thank you

Issue #170, opened 2021-08-06
Support irregular indentation


lukas-reineke:
When using indent guides for Haskell code -- see the image.

![image](https://user-images.githubusercontent.com/15617291/128443501-5f24189b-0d2d-451b-bb56-3e58f1e91b1c.png)

The only non-default settings here are `g:indent_blankline_char`, `g:indent_blankline_buftype_exclude` and `g:indent_blankline_show_trailing_blankline_indent = v:false`.

I believe this happens because this plugin simply places an indentation guide every `&shiftwidth` columns. But, as the image shows, this is not enough for typical Haskell indentation, since it can start at basically any column.

Perhaps we could look into placing the indent guide on the first non-whitespace column above? Or maybe somehow use tree-sitter to try to solve this? (note: using `g:indent_blankline_use_treesitter = 1` didn't solve this, as I understand all it does is highlight the current indentation level).

_Originally posted by @dccsillag in https://github.com/lukas-reineke/indent-blankline.nvim/issues/153#issuecomment-893943554_

lukas-reineke:
I made this a new issue.

It's an interesting problem. I believe it could be solvable, but it's not easy.

dccsillag:
Maybe it's worth documenting another instance of this problem (again exemplified in Haskell code): multiple indent guides where there should only be one.

![image](https://user-images.githubusercontent.com/15617291/128447810-cbc3cb82-4da4-4f7c-be92-788eac548e87.png)

lukas-reineke:
do you have examples from other editors how it _should_ look like?

dccsillag:
I have an idea in my head on how it should look like, but I feel like I've also seen some editor out there do it properly. I'll look for it.

dccsillag:
Update: Just found https://github.com/DarthFennec/highlight-indent-guides, which implements this stuff in Emacs.

1478zhcy:
Obviously, show current context indent works well, could fully use treesitter implement all feature of this plugin?

lukas-reineke:
It is on the list of things I want to do in the next version.
But I have much less time these days, I don't know when I'll work on it.

lukas-reineke:
Supported in version 3

PR #204, opened 2021-09-03
Add option max_indent_increase

zeertzjq:
This adds an option `max_indent_increase`, which specifies the maximum indent level increase compared to the last normally indented line that make a line a normally indented line. For abnormally indented lines, only the normal indent level is shown. This makes aligned comments look better.

### DIFF
--- a/lua/indent_blankline/init.lua
+++ b/lua/indent_blankline/init.lua
@@ -54,6 +54,8 @@ M.setup = function(options)
         o(options.buftype_exclude, vim.g.indent_blankline_buftype_exclude, vim.g.indentLine_bufTypeExclude, {})
     vim.g.indent_blankline_viewport_buffer = o(options.viewport_buffer, vim.g.indent_blankline_viewport_buffer, 10)
     vim.g.indent_blankline_use_treesitter = o(options.use_treesitter, vim.g.indent_blankline_use_treesitter, false)
+    vim.g.indent_blankline_max_indent_increase =
+        o(options.max_indent_increase, vim.g.indent_blankline_max_indent_increase, vim.g.indent_blankline_indent_level)
     vim.g.indent_blankline_show_first_indent_level =
         o(options.show_first_indent_level, vim.g.indent_blankline_show_first_indent_level, true)
     vim.g.indent_blankline_show_trailing_blankline_indent =
@@ -160,6 +162,7 @@ local refresh = function()
     end
 
     local max_indent_level = v("indent_blankline_indent_level")
+    local max_indent_increase = v("indent_blankline_max_indent_increase")
     local expandtab = vim.bo.expandtab
     local use_ts_indent = v("indent_blankline_use_treesitter") and ts_status and ts_query.has_indents(vim.bo.filetype)
     local first_indent = v("indent_blankline_show_first_indent_level")
@@ -177,10 +180,18 @@ local refresh = function()
         context_status, context_start, context_end = utils.get_current_context(v("indent_blankline_context_patterns"))
     end
 
-    local get_virtual_text = function(indent, extra, blankline, context_active, context_indent, virtual_string)
+    local get_virtual_text = function(
+        indent,
+        extra,
+        blankline,
+        context_active,
+        context_indent,
+        prev_indent,
+        virtual_string)
         local virtual_text = {}
         local current_left_offset = left_offset
-        for i = 1, math.min(math.max(indent, 0), max_indent_level) do
+        local local_max_indent_level = math.min(max_indent_level, prev_indent + max_indent_increase)
+        for i = 1, math.min(math.max(indent, 0), local_max_indent_level) do
             local space_count = shiftwidth
             local context = context_active and context_indent == i
             if i ~= 1 or first_indent then
@@ -262,7 +273,10 @@ local refresh = function()
             end
         end
 
-        if ((blankline or extra) and trail_indent) and (first_indent or #virtual_text > 0) and current_left_offset < 1 then
+        if
+            ((blankline or extra) and trail_indent) and (first_indent or #virtual_text > 0) and current_left_offset < 1 and
+                indent < local_max_indent_level
+         then
             local index = math.ceil(#virtual_text / 2) + 1
             table.insert(
                 virtual_text,
@@ -292,6 +306,7 @@ local refresh = function()
         return virtual_text
     end
 
+    local prev_indent
     local next_indent
     local next_extra
     local empty_line_counter = 0
@@ -320,11 +335,19 @@ local refresh = function()
                                 end
                                 indent = indent / shiftwidth
                                 if offset + i == context_start then
-                                    context_indent = (indent or 0) + 1
+                                    context_indent = indent + 1
                                 end
 
                                 local virtual_text =
-                                    get_virtual_text(indent, false, blankline, context_active, context_indent, {})
+                                    get_virtual_text(
+                                    indent,
+                                    false,
+                                    blankline,
+                                    context_active,
+                                    context_indent,
+                                    max_indent_level,
+                                    {}
+                                )
                                 utils.clear_line_indent(bufnr, i + offset)
                                 xpcall(
                                     vim.api.nvim_buf_set_extmark,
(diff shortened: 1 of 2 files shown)


lukas-reineke:
I think you are misunderstanding #170 
That ticket is about showing indent lines on each indentation level.
Like this

zeertzjq:
Oh, this is unrelated to #170 then.

zeertzjq:
I mainly create this to make aligned comments look less ugly.

Original behavior:

Screenshot_20210903_121827

With `g:indent_blankline_indent_increment = 1`:

Screenshot_20210903_121847

lukas-reineke:
Okay I get what you are trying to do :+1: 
I'm okay with adding this, but your implementation is too complicated. You can write this a lot easier.
All you need to do is update `indent` before you call `get_virtual_text`.

This should be enough
```lua
if indent > last_indent + max_indent_increase then
    indent = last_indent + max_indent_increase
else
    last_indent = indent
end
```

zeertzjq:
The first commit is unrelated to the new option. It prevent showing extra chars if maximum indent level is already reached.

zeertzjq:
It seems to stop working.

lua/indent_blankline/init.lua
```diff
@@ -143,6 +145,11 @@ local refresh = function()
     }
 
     local max_indent_level = v("indent_blankline_indent_level")
+    local max_indent_increase = v("indent_blankline_max_indent_increase")
+    if max_indent_increase <= 0 then
+        max_indent_increase = max_indent_level
+    end
```
lukas-reineke, line 151: instead of this, can you just make the default for `indent_blankline_max_indent_increase = indent_blankline_indent_level`

lua/indent_blankline/init.lua
```diff
@@ -143,6 +145,11 @@ local refresh = function()
...
+    local last_normal_indent = -1
```
lukas-reineke, line 152: Can you rename this to `previous_indent` and move it to line 288, above `next_indent`

Also, just start with `0`, then you can remove the `previous_indent ~= -1` checks

lua/indent_blankline/init.lua
```diff
@@ -352,6 +368,15 @@ local refresh = function()
                     if offset + i == context_start then
                         context_indent = (indent or 0) + 1
                     end
+                    if
+                        last_normal_indent ~= -1 and
+                            indent + (extra and 1 or 0) > last_normal_indent + max_indent_increase
+                     then
+                        indent = last_normal_indent + max_indent_increase
+                        extra = false
```
lukas-reineke, line 376: The `extra` check should only happen when `trail_indent` is true.
Also, `blankline` here works the same as `extra`. Either of them adds +1

lukas-reineke (changes requested):

lua/indent_blankline/init.lua
zeertzjq: > just start with `0`

This will cause problem if the first line is indented.

lua/indent_blankline/init.lua
zeertzjq: I am thinking about reverting to <https://github.com/lukas-reineke/indent-blankline.nvim/commit/de4b52818fbcadae9b3fac65cf909beed475ee0e> and renaming the variables.

lua/indent_blankline/init.lua
zeertzjq: Won't this require users to change their config if they use a non-default `indent_level` value?

lukas-reineke: Same with what you have currently too. Setting the default as `0` and then changing it here, or setting the default as `max_indent_level` directly, doesn't make a difference. Just makes the code easier.

lua/indent_blankline/init.lua
lukas-reineke: What problem exactly? Maybe I'm missing it, but I don't see anything wrong

zeertzjq: If the whole viewport is indented, all lines are treated as abnormal lines.

lua/indent_blankline/init.lua
```diff
@@ -160,10 +166,13 @@ local refresh = function()
         context_status, context_start, context_end = utils.get_current_context(v("indent_blankline_context_patterns"))
     end
 
+    local prev_indent = nil
+
     local get_virtual_text = function(indent, extra, blankline, context_active, context_indent, virtual_string)
         local virtual_text = {}
         local current_left_offset = left_offset
-        for i = 1, math.min(math.max(indent, 0), max_indent_level) do
+        local local_max_indent_level = math.min(max_indent_level, prev_indent + max_indent_increase)
```
lukas-reineke, line 174: `get_virtual_text` can't use dynamic variables from outside. It runs async. This is not guaranteed to work.
Revert it back to how you had it before. Update the indent value before you call `get_virtual_text`

zeertzjq: Maybe I can just pass prev_indent as a parameter?

lua/indent_blankline/init.lua
lukas-reineke: :+1: yeah you are right

lua/indent_blankline/init.lua
lukas-reineke: that works for now

really it shouldn't be the responsibility of `get_virtual_text`, but a lot of stuff in there shouldn't. I have to clean up the code at some point.

lua/indent_blankline/init.lua
```diff
@@ -275,6 +289,7 @@ local refresh = function()
         return virtual_text
     end
 
+    local prev_indent = nil
```
lukas-reineke, line 292: just to stay consistent with the other variables
```suggestion
    local prev_indent
```

lua/indent_blankline/init.lua
```diff
@@ -352,14 +379,25 @@ local refresh = function()
                     if offset + i == context_start then
                         context_indent = (indent or 0) + 1
                     end
+                    if not prev_indent or indent + (extra and 1 or 0) <= prev_indent + max_indent_increase then
+                        prev_indent = indent
```
lukas-reineke, line 383: the other code here assumes `indent` could be undefined.
I actually think this can't happen anymore, but I am not 100% sure :sweat_smile: just to be safe, do `(indent or 0)`

lua/indent_blankline/init.lua
```diff
@@ -298,16 +313,28 @@ local refresh = function()
                             function()
                                 local indent = ts_indent.get_indent(i + offset)
                                 if not indent or indent == 0 then
+                                    prev_indent = 0
                                     utils.clear_line_indent(bufnr, i + offset)
                                     return
                                 end
                                 indent = indent / shiftwidth
+                                if not prev_indent or indent <= prev_indent + max_indent_increase then
+                                    prev_indent = indent
```
lukas-reineke, line 322: here as well `(indent or 0)`

lua/indent_blankline/init.lua
zeertzjq: I'll just move it after the zero check.

lua/indent_blankline/init.lua
```diff
@@ -354,12 +381,25 @@ local refresh = function()
                     end
 
                     if (not indent or indent == 0) and #virtual_string == 0 then
+                        prev_indent = 0
                         vim.schedule_wrap(utils.clear_line_indent)(bufnr, i + offset)
                         return async:close()
                     end
 
+                    if not prev_indent or indent + (extra and 1 or 0) <= prev_indent + max_indent_increase then
+                        prev_indent = indent
```
lukas-reineke, line 390: This still doesn't handle `trail_indent`
If it's off, all lines that take the indent from the previous line need to be indent one level less

```suggestion
                        prev_indent = indent - utils._if(trail_indent, 1, 0)
```

zeertzjq: I think this should be done in blank lines only?

zeertzjq: `trail_indent` seems to only affect what `get_virtual_text` returns. It doesn't affect the value of `indent`. I tried to add this and it broke.

lukas-reineke: sorry I got the `0` and `1` switched before. Fixed it now.

With `trail_indent` off, it strips one indent level from blanklines and the extra indent.
This is how it should look like

on:
Screen Capture_select-area_20210903195559

off:
Screen Capture_select-area_20210903195612

zeertzjq: After adding this, with `max_indent_increase = 1` only the first indent level is shown.

zeertzjq: Oh, I think I understood what you meant.

zeertzjq:
There still seems to be some problems with blank line.

zeertzjq:
> There still seems to be some problems with blank line.

Maybe there is a racing condition?

zeertzjq:
I guess Treesitter doesn't need `prev_mode` because it understands the syntax.

lukas-reineke:
Code looks good now :+1: 
you have to rebase against develop please.

I want to test this for a bit longer before I merge it. But I think it should be okay.
Thank you

lukas-reineke (approved):

Issue #254, opened 2021-10-08
won't display space for blankline


stumash:
Here's my config
```vim
set lcs=space:·,tab:→\ ,eol:↵
hi NonText ctermfg=0 guifg=#282828
hi SpecialKey ctermfg=0 guifg=#222222
set list
```

```lua
vim.cmd [[highlight IndentBlanklineIndent1 guibg=#1f1f1f guifg=#383838 gui=nocombine]]
vim.cmd [[highlight IndentBlanklineIndent2 guibg=#1a1a1a guifg=#383838 gui=nocombine]]

require("indent_blankline").setup {
    char = "·",
    char_highlight_list = {
        "IndentBlanklineIndent1",
        "IndentBlanklineIndent2",
    },
    space_char_blankline = " ",
    space_char_highlight_list = {
        "IndentBlanklineIndent1",
        "IndentBlanklineIndent2",
    },
    show_trailing_blankline_indent = false,
}
```

Here's what it looks like
![image](https://user-images.githubusercontent.com/13181078/136594596-38d289a6-92bc-4eaf-94ea-6ca5bd162f6d.png)

It's still displaying a dot (weirdly, not two) on the blank lines, even though I set the `space_char_blankline = " "`

Any idea what's going on?

stumash:
it's not a bad feature request come to think of it

Daxtorim:
I'm not sure if a `char_blankline` would be particularly useful as a user defined setting. To me it looks like what you really want is to disable the indent `char` completely and only rely on the highlighting of the background. 

What I mean is something so you can set `char = ""` and it will just not insert any indent `char` at all and just use the underlying whitespace (or lack thereof) to display `listchars`. I'll do some testing...

PR #255, opened 2021-10-09
 feature: allow empty string for indent char

Daxtorim:
Addresses #254 

1. Check if `char` variable is set and only insert indent character when
   not empty. Insert additional character from real underlying
   whitespace instead.

When setting the `char = ""` before, the resulting virtual text would
have been too short since it won't insert any character as a
replacement.  This also affected the highlighting of the virtual text.
Additionally when no indent char was desired, the only way to "ignore"
it was to set `char` to a space. This, however, doesn't look good when
also displaying `listchars` because it will insert an empty space even
when the user has set a different character to be displayed in place of
a space.

<s>2. Use `options.space_char_highlight_list` before sourcing
   `vim.g.indent_blankline_space_char_blankline_highlight_list`.

For some reason `utils.first_not_nil()` won't get past the global vim
variable even if it isn't set and therefore never gets to the option
that was passed to the setup function for fallback. If the user passed
in a list for one of the settings it is save to assume they would have
also passed a list for the other one if they wanted them to be unique
instead of setting a mish-mash of function arguments and global vim
variables.</s>

3. Change README.md to reflect new option when sensible.

### DIFF
--- a/README.md
+++ b/README.md
@@ -123,7 +123,7 @@ vim.cmd [[highlight IndentBlanklineIndent1 guibg=#1f1f1f gui=nocombine]]
 vim.cmd [[highlight IndentBlanklineIndent2 guibg=#1a1a1a gui=nocombine]]
 
 require("indent_blankline").setup {
-    char = " ",
+    char = "",
     char_highlight_list = {
         "IndentBlanklineIndent1",
         "IndentBlanklineIndent2",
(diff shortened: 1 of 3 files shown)


stumash:
just wanted to say thanks for being good maintainers. send me link to buy you a coffee

lukas-reineke:
> just wanted to say thanks for being good maintainers. send me link to buy you a coffee

Thank you, I really appreciate that. You can sponsor me on GitHub https://github.com/sponsors/lukas-reineke

Daxtorim:
Fixed support for the end of line character and removed unnecessary changes.

PR #267, opened 2021-10-16
perf(scrolling): optimize indent line refresh during scrolling

jbarap:
Hello!

Big fan of the plugin, but there's something that has always bothered me, and that is the performance in big files.

I looked around in the code and noticed that, as far as I'm aware, the indent lines get refreshed every time the screen scrolls, which in my slow computer leads to really choppy scrolling, whether while holding down `j` and `k` or with plugins like Neoscroll.

Here's a proposed optimization: when you scroll through a file the indents won't have to be updated, because the content itself hasn't changed, so why not draw the indents on zones where there's no indent yet, and only re-draw the indents when you encounter another empty zone.

I called the empty zones "ranges", and what this PR does is keep track of the parts of the file that already have indents, and expand those ranges any time we draw more intent lines on the screen.

I wanted to know if you're interested in implementing this, or if you have better ideas as to how to optimize this. I have tested some common scenarios, but of course there are probably going to be some edge cases which I haven't seen.

Here's a profile that I ran with the built in `:profile file *` and `:profile func *`, on both of these I scrolled through the file `src/nvim/eval/funcs.c` in the neovim repo, holding down the `j` key from line `0` to line `2000`:

Current implementation:
image

After the PR:
image

It's roughly twice as fast, and it feels considerably smoother.

### DIFF
--- a/lua/indent_blankline/init.lua
+++ b/lua/indent_blankline/init.lua
@@ -166,15 +166,15 @@ local refresh = function(scroll)
         vim.b.__indent_blankline_active = true
     end
 
-    local win_start = vim.fn.line("w0")
-    local win_end = vim.fn.line("w$")
+    local win_start = vim.fn.line "w0"
+    local win_end = vim.fn.line "w$"
     local offset = math.max(win_start - 1 - v "indent_blankline_viewport_buffer", 0)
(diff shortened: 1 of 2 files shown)


lukas-reineke:
Thanks
I'm currently on vacation, I'll take a look when I'm back

JA-Bar:
Sure, enjoy your time off!

PR #274, opened 2021-10-23
feature: support a different character for the current context indent line

sthendev:
new options:
- `g:indent_blankline_context_char`
- `g:indent_blankline_context_char_list`

When specified, `context_char` will take precedence over `char` for the current context indent line.


lukas-reineke:
Code is all good now :+1: just fix formatting with stylelua please

PR #297, opened 2021-11-04
fix(refresh): fix range updates when scrolling

JA-Bar:
Fixes bug introduced by 1b554c6183e54b3f1d7266ecc9d3be1aa1a58349, where the ranges table wasn't getting updated at all, effectively nullifying the scrolling optimization.

For some reason when you directly assign a value to a key in `vim.b.my_table`, the table doesn't actually update, so a solution is to assign the table to a local variable, change the key in that local variable, then assign the variable to `vim.b.my_table`. I don't know if it's related to the way neovim accesses values from lua or why this happens, but if you have a better solution let me know.

### DIFF
--- a/lua/indent_blankline/init.lua
+++ b/lua/indent_blankline/init.lua
@@ -184,6 +184,8 @@ local refresh = function(scroll)
     end
 
     if scroll then
+        local updated_range
+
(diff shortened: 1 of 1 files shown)


lukas-reineke (approved):
thanks :rocket:

PR #342, opened 2021-12-30
doc: added 'excluding filetypes section'

lolix-notepad:
I hope this can be useful, I'm open to any suggestions.


lukas-reineke:
Thank you, but everything is already documented in the vim docs `:help indent-blankline`
I do not want to maintain the settings in the readme as well.

PR #437, opened 2022-06-12
Avoid errors on cmdline window

delphinus:
With this plugin, Neovim reports errors when the cursor goes into the cmdline window.

You can see errors below with `q:` in the normal mode.

```
Vim(lua):E5108: Error executing lua .../indent-blankline.nvim/lua/indent_blankline/commands.lua:7: Vim(windo):E11: Invalid in command-line
 window; <CR> executes, CTRL-C quits: noautocmd windo lua require(\"indent_blankline\").refresh(false)
```

### DIFF
--- a/lua/indent_blankline/commands.lua
+++ b/lua/indent_blankline/commands.lua
@@ -4,8 +4,8 @@ M.refresh = function(bang, scroll)
     scroll = scroll or false
     if bang then
         local win = vim.api.nvim_get_current_win()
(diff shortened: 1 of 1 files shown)


lukas-reineke:
Thank you, but I don't think this error should be removed.
You can't run global refresh and enable from the command line window. Users should get an error if they try, instead of just nothing happening.

> Neovim reports errors when the cursor goes into the cmdline window.

If this happens every time, you have misconfigured something.

delphinus:
Oh, OK. I'll check my vimrc's.

PR #447, opened 2022-06-23
Simplified if null then `value`

asteroidalaz:


lukas-reineke:
Thanks,
Can you please rebase against develop and format the code with stylua

PR #453, opened 2022-07-11
Add `CursorMovedI` event to `IndentBlanklineRefresh` autocmd

utilyre:
It is obvious that users shouldn't move around in insert mode, however
in some scenarios the cursor moves out of the current context in insert
mode.

For example let's say we are editing something like this

```c
if (myBool) {

} // <-- cursor is on this line
```

if the user presses `o` the cursor moves out of the if statement context
with the `CursorMovedI` event being called instead of `CursorMoved`.

Or if the user has an autopair plugin by pressing `{<cr>` that plugin
*potentially* creates a new context and moves the cursor inside that
context in insert mode.

### DIFF
--- a/doc/indent_blankline.txt
+++ b/doc/indent_blankline.txt
@@ -467,7 +467,8 @@ g:indent_blankline_show_current_context *g:indent_blankline_show_current_context
     Note: Requires https://github.com/nvim-treesitter/nvim-treesitter to be
           installed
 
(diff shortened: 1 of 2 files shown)


lukas-reineke:
Thanks, looks good.
Can you open the PR against develop?

lukas-reineke:
You have to rebase against develop.
Should be just `git rebase develop` and then force push with `git push --force`

PR #456, opened 2022-07-13
Add `refresh_events` option

utilyre:
This is a better approach of the previous pr I made (#453). This time instead of hard coding the events I decided to let *the end user* choose what events should run the `IndentBlanklineRefresh` command.

Option name: `refresh_events`
Default value: `{ "CursorMoved" }`

As I mentioned in the commit message please add this to the help tag since I don't know much about that. Thanks!

### DIFF
--- a/lua/indent_blankline/init.lua
+++ b/lua/indent_blankline/init.lua
@@ -183,7 +183,7 @@ M.setup = function(options)
         vim.cmd [[
             augroup IndentBlanklineContextAutogroup
                 autocmd!
-                autocmd CursorMoved * IndentBlanklineRefresh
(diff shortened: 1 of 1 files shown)


lukas-reineke:
I get what you are going for, but this would be a more complicated change.
This is not the only place autocmds are set. They are different based on if context is on or not.

I also don't think it is that necessary to have a configuration option for this. Users can just overwrite the autogroup if they want to customize the events.

But I'd still like to merge your first version, just fixing the default autocmd.

lukas-reineke (approved):
Thanks

The failing check is unrelated, I'll fix that in develop

PR #463, opened 2022-07-31
fix help tag

ryota2357:
Do not use \*~\* for if.

PR #475, opened 2022-08-28
Remove if helptags

yuki-yano:
Because the doc contains an if tag, it conflicts with the if in the vim doc and causes an error.
The error in dein is as follows.
```
[dein] Error generating helptags:
[dein] Vim(helptags):E154: Duplicate tag "if" in file /Users/yuki-yano/.vim/bundle/.cache/init.vim/.dein/doc/indent_blankline.txt
[dein] function dein#recache_runtimepath[1]..dein#install#_recache_runtimepath[31]..<SNR>175_helptags, line 10
```


lukas-reineke:
Thanks, but this has already been fixed in #463

PR #484, opened 2022-09-30
Left aligned vertical bar character

sassanh:
With this vertical line character we probably won't need `modify_font` as it is already on the left edge of the character box.

Example usage in iTerm2:

image

### DIFF
--- a/lua/indent_blankline/init.lua
+++ b/lua/indent_blankline/init.lua
@@ -24,7 +24,7 @@ M.setup = function(options)
 
     local o = utils.first_not_nil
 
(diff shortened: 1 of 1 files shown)


lukas-reineke:
Thank you,

But the character does not affect `modify_font`. `modify_font` controls the thickness and vertical position of the underline, it is not to align horizontally.
But I am using the left aligned character in the screenshot, it should be in the example config for context in the readme.
I don't want to change the default character, that would be a pretty big breaking change for everyone. Some terminals/fonts might not play well with that character.

Can you change the PR to only update the example config in the readme?

HawkinsT:
I'd suggest giving u258e as the example instead of u258f as it's slightly thicker, mirroring the appearance of the default character (u2502), only left aligned.

PR #507, opened 2022-11-28
doc: in README.md examples use `vim.api.nvim_set_hl` instead of `vim.cmd([[highlight ...]])`

Senpai-10:
no need to use vimscript for this :)

### DIFF
--- a/README.md
+++ b/README.md
@@ -88,12 +88,12 @@ require("indent_blankline").setup {
 
 ```lua
 vim.opt.termguicolors = true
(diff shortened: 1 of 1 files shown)


lukas-reineke:
This plugin is backwards compatible until Neovim 0.5 (for the most part)
`nvim_set_hl` is newer than that.

I want to deprecate older Neovim versions in v3. Please leave the PR open, I'll merge it then.

lukas-reineke:
Code in v3 changed too much, I'll close this

PR #538, opened 2023-02-05
Improved detection for present highlight groups

nenikitov:
Before, it used to check if `fg`, `bg`, or `sp` are present, and if no created a highlight group. It would make impossible for a cterm only color scheme to redefine these highlight groups.

Now, it simply checks if the highlight group was already created before.


lukas-reineke:
Checking if the highlight group was created before only works once. The highlight groups get fixed every time a new color scheme is set.

PR #543, opened 2023-02-17
add `match`  to context patterns

steven-omaha:
Currently the context is shown like this. The `match` block is not highlighted.

without

With the patch, the `match` block is shown as context.

with


lukas-reineke (approved):
thank you

PR #564, opened 2023-04-23
added treesitter support for zig

chrboesch:
The treesitter commands for Zig differ, which is why context highlighting did not work. I have added the necessary commands in this PR.


lukas-reineke (approved):
Thanks
and sorry for the delay

PR #570, opened 2023-05-03
doc: Fix man page formatting

jrwrigh:
These extra blanklines were messing up the format highlighting:
image


lukas-reineke:
I think having the newlines is correct :thinking:

jrwrigh:
I can't find many examples out there, but [mason](https://github.com/williamboman/mason.nvim/blob/main/doc/mason.txt) has no newlines between `-----` and the next line. But also, those are used for subsection headers rather than on every setting. So I think those `-----` should probably be removed all together?

lukas-reineke (approved):
actually looking at the treesitter code this looks correct :+1:

PR #578, opened 2023-05-29
Clear indent on topline when skipcol is non-zero

luukvbaal:
Fixes the situation where the start of a wrapped topline is hidden behind the indentation markers:
image
image

This has become more relevant with the addition of 'smoothscroll'.

Also remove a call to `vim.fn.line()` and use the result of `vim.fn.winsaveview().topline` for `win_start` instead.

### DIFF
--- a/lua/indent_blankline/init.lua
+++ b/lua/indent_blankline/init.lua
@@ -181,10 +181,10 @@ local refresh = function(scroll)
         vim.b.__indent_blankline_active = true
     end
 
-    local win_start = vim.fn.line "w0"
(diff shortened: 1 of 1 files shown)


luukvbaal:
> Need to do something else for that.

Not sure what TBH, need to clear/apply indentation markers for the topline if skipcol changed. Don't really see a way to do that without nullifying the early `need_to_update` `scroll` return or substantial refactoring.

Perhaps see this as an issue report.

luukvbaal:
Actually this should probably be fixed in the API: https://github.com/neovim/neovim/issues/23775. Closing.

PR #582, opened 2023-06-24
docs: Make custom background highlighting example automatically set color

brycepg:
based off of existing background color

### DIFF
--- a/README.md
+++ b/README.md
@@ -114,8 +114,38 @@ require("indent_blankline").setup {
 
 ```lua
 vim.opt.termguicolors = true
-vim.cmd [[highlight IndentBlanklineIndent1 guibg=#1f1f1f gui=nocombine]]
-vim.cmd [[highlight IndentBlanklineIndent2 guibg=#1a1a1a gui=nocombine]]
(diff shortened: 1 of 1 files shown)


brycepg:
Let me know if you think this is a good idea. If you do then maybe I can include it as an option instead of as docs with some pointers

brycepg:
I'm also sure there is a way to trigger the adjustment of colors when the bgcolor or colorscheme changes as well

brycepg:
Here's what it looks like with my colorscheme:

indent-colors-3

indent-colors-3-wombat

lukas-reineke:
Thank you, but I don't think this is an improvement for the documentation. It is much harder to understand what's going on.

And it's also out of scope to add as an option. Getting this color is a one time thing, it shouldn't be computed every time Neovim starts.

brycepg:
Whatever

brycepg:
Whatever will we do if the computer actually computes things

brycepg:
Thanks for letting me know. I think I'm going to fork and change scope

PR #588, opened 2023-07-05
Fix for Ruby 'unless' blocks

tradiff:
Although its usefulness may be subject to debate, Ruby includes an `unless` block. This block isn't currently recognized by indent-blankline's `show_current_context` feature.

- Add support for ruby's `unless` blocks
- Fix some inconsistent whitespace

Here is a before and after screenshot.

## Before:
The cursor is on line 8.
The context line is on lines 4-11 instead of the expected 5-8.
image

## After:
The cursor is on line 8.
The context line is on lines 5-8, which is what one would expect.
image

### DIFF
--- a/lua/indent_blankline/init.lua
+++ b/lua/indent_blankline/init.lua
@@ -118,14 +118,15 @@ M.setup = function(options)
         "table",
         "tuple",
         "do_block",
-		"Block",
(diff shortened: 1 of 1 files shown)


lukas-reineke (approved):
Thanks

PR #608, opened 2023-08-21
add checks for cterm highlights before highlights reset

### LINKED ISSUE #606: highlight groups get reset if I only set `ctermfg`
If in my colorscheme I set, for example:
```lua
vim.api.nvim_set_hl(0, 'IndentBlanklineChar', { ctermfg = 8 })
```
with `vim.opt.termguicolors = false`, the highlight get's reset by this plugin in `require("indent_blankline.utils").reset_highlights()` function.

I've tried to debug this and I'm really not sure why. It seems like some highlight groups that only have `ctermfg` get defined later? Which doesn't make any sense. Specifically this seems to be happening with `IndentBlanklineContextChar` group – in `reset_highlights` function, this group is empty, even though I define it in my colorscheme.
This seems to be a neovim issue.
For all highlight groups, `vim.fn.synIDattr(current_highlight, "fg")`, doesn't return anything after `colorscheme` call from my vimrc. But after `colorscheme` call from within neovim, everything works as it should. I'm confused.

Anyways, I believe there is an easy solution for this. The `reset_highlights` function could only be called on `ColorSchemePre` event. Then there would also be no need to check if the highlights are already defined, the function would just reset them and let the colorscheme override any of them if needed. I could make a PR.

ramojus:
fixes #606

This could be a neovim issue. Quote from help (note the last sentence):
```
synIDattr({synID}, {what} [, {mode}])			*synIDattr()*
		The result is a String, which is the {what} attribute of
		syntax ID {synID}.  This can be used to obtain information
		about a syntax item.
		{mode} can be "gui" or "cterm", to get the attributes
		for that mode.  When {mode} is omitted, or an invalid value is
		used, the attributes for the currently active highlighting are
		used (GUI or cterm).
```
But it seems to only detect that I'm using cterm highlighting when I change colorscheme from within neovim, but not when I set colorscheme from my vimrc. Maybe the "currently active highlighting" is only defined after neovim launches.

Anyways, this seems like a simple fix.

### DIFF
--- a/lua/indent_blankline/utils.lua
+++ b/lua/indent_blankline/utils.lua
@@ -250,6 +250,8 @@ M.reset_highlights = function()
             vim.fn.synIDattr(current_highlight, "fg") == ""
             and vim.fn.synIDattr(current_highlight, "bg") == ""
             and vim.fn.synIDattr(current_highlight, "sp") == ""
+            and vim.fn.synIDattr(current_highlight, "fg", "cterm") == ""
+            and vim.fn.synIDattr(current_highlight, "bg", "cterm") == ""
         then
             if highlight_name == "IndentBlanklineContextStart" then
                 vim.cmd(


lukas-reineke (approved):
Thanks :rocket: 
This logic will be better in v3, for now this is fine

PR #612, opened 2023-09-04
Indent blankline v3

lukas-reineke:
See https://github.com/lukas-reineke/indent-blankline.nvim/wiki/Migrate-to-version-3


lua/ibl/init.lua
```diff
@@ -0,0 +1,381 @@
...
+local debounced_refresh = setmetatable({
+    timers = {},
+}, {
+    ---@param bufnr number
+    __call = function(self, bufnr)
+        bufnr = utils.get_bufnr(bufnr)
+        local config = conf.get_config(bufnr)
+        if not self.timers[bufnr] or self.timers[bufnr]:is_closing() then
+            ---@diagnostic disable-next-line:undefined-field
+            self.timers[bufnr] = vim.uv.new_timer()
```
kuznetsss, line 120: I'm getting an error that `vim.uv` is `nil`. Should it be `vim.loop`?
I'm using neovim v0.9.1.

dtomvan: Yes, that is virtually the same thing. We could check nvim version (with features or the new lua api, although that wil ofcourse not work for older versions) wether we should use vim.uv or vim.loop

majamin: vim.loop / vim.uv feature discussion https://github.com/neovim/neovim/pull/22846/

dtomvan: Ah, my bad, vim.loop always exists. We should probably keep that then, for simplicity's sake. Even though the issue you mentioned specifies that we should probably not do that, but in the spirit of backwards compatibility I think we should as long as it doesn't become a mess.

lua/ibl/autocmds.lua
```diff
@@ -0,0 +1,58 @@
+local highlights = require "ibl.highlights"
+local M = {}
+
+M.setup = function()
+    local group = vim.api.nvim_create_augroup("IndentBlankline", {})
```
wookayin, line 5: `{ clear = true }` needed

lukas-reineke: `clear` defaults to true

lua/ibl/init.lua
lukas-reineke: should be fixed now, thanks

simonmandlik:
Hi, I'm getting the following selection behavior with default setup:

`vip`:

image

With master, I get

image

Is this intended?

lua/ibl/autocmds.lua
```diff
@@ -0,0 +1,61 @@
...
+        end,
+    })
+    vim.api.nvim_create_autocmd("ColorScheme", {
+        group = group,
+        pattern = "*",
+        callback = function()
+            highlights.setup()
+            ibl.refresh_all()
+        end,
+    })
```
simonmandlik, line 58: Isn't this a bit strict?

I cannot for example use a custom hlg, which is not set by the colorscheme itself.

`minimal.lua`:
```lua
-- DO NOT change the paths and don't remove the colorscheme
local root = vim.fn.fnamemodify("./.repro", ":p")

-- set stdpaths to use .repro
for _, name in ipairs({ "config", "data", "state", "cache" }) do
  vim.env[("XDG_%s_HOME"):format(name:upper())] = root .. "/" .. name
end

-- bootstrap lazy
local lazypath = root .. "/plugins/lazy.nvim"
if not vim.loop.fs_stat(lazypath) then
  vim.fn.system({ "git", "clone", "--filter=blob:none", "https://github.com/folke/lazy.nvim.git", lazypath, })
end
vim.opt.runtimepath:prepend(lazypath)

vim.cmd [[hi MyHLG guifg=red]]

-- install plugins
local plugins = {
    {
        "lukas-reineke/indent-blankline.nvim",
        branch = "v3",
        config = function()
            require("ibl").setup {
                indent = { highlight = "MyHLG" },
            }
        end
    },
}
require("lazy").setup(plugins, {
  root = root .. "/plugins",
})

vim.cmd [[set termguicolors]]
```

Now when I run for example `colorscheme default`, it clears `MyHLG`, and when the autocommand runs it errors

lukas-reineke: You need to use the `HIGHLIGHT_SETUP` hook, it's explained in the readme and the docs

https://github.com/lukas-reineke/indent-blankline.nvim/tree/v3#multiple-indent-colorsh

For the visual selection, I can't reproduce that. Please open an issue with steps to get there.

simonmandlik: Hmm, yes you are right. I just read `:h ibl.setup()` and  scrolled to `config.indent`. Out of curiosity — is there a reason why hooks are necessary here? None of other plugins I use with custom hl groups require that (`tzachar/highlight-undo.nvim`, `winston0410/range-highlight.nvim`, `folke/flash.nvim` to name a few), and their config is more concise. 

About the highlighting — I'm not sure it is an "issue" per see, but side effect of how highlights are handled as well. Try this

```lua
local root = vim.fn.fnamemodify("./.repro", ":p")
for _, name in ipairs({ "config", "data", "state", "cache" }) do
  vim.env[("XDG_%s_HOME"):format(name:upper())] = root .. "/" .. name
end
local lazypath = root .. "/plugins/lazy.nvim"
if not vim.loop.fs_stat(lazypath) then
  vim.fn.system({ "git", "clone", "--filter=blob:none", "https://github.com/folke/lazy.nvim.git", lazypath, })
end
vim.opt.runtimepath:prepend(lazypath)

local plugins = {
    "RRethy/nvim-base16",
    {
        "lukas-reineke/indent-blankline.nvim",
        branch = "v3",
        config = function()
            require("ibl").setup { }
        end
    },
}

require("lazy").setup(plugins, {
    root = root .. "/plugins",
})

vim.cmd [[set termguicolors]]
vim.cmd [[colorscheme base16-dirtysea]]
vim.opt.background = "light"
vim.api.nvim_set_hl(0, "NonText",  { fg = "Gray", bg = "bg" })
```

Anyway, thanks for the hard work on this plugin!

lukas-reineke: I agree it's not ideal, but I don't see a better way. Because there can be different colors for the indent guides, the scope, and the whitespace, I need to generate all combinations for all cases where they are used.

Otherwise, to have 3 colors each, the user would need to manually make 42 highlight groups. That would be much less user-friendly.
Even with just one color each, I still need to create 7 highlight groups.

PR #618, opened 2023-09-04
docs: fix typos

MariaSolOs:
Fixes a couple of typos I found while reading the new docs.

Also removes a duplicate tag for `ibl.config.scope.exclude` that causes an error when building the docs with `lazy.nvim`.


lukas-reineke (approved):
Thank you :rocket:

PR #659, opened 2023-09-29
fix: add support for Neovim v0.7+

### LINKED ISSUE #658: Hard requirement for Neovim v0.9+
### Problem

Currently the new v3 is using the very recently added `vim.api.nvim_get_hl`. This was added in Neovim v0.9 and many people still use older versions of Neovim which only have `vim.api.nvim_get_hl_by_name`. Could this be added into the code to support older versions of Neovim? Most plugins support Neovim v0.7+ rather than requiring only the most up to date release.

### Steps to reproduce

Just start neovim with version 0.8 with any amount of minimal configuration that just adds the plugin:

```sh
nvim --clean -u min-init.lua
```

### Expected behavior

I would expect support for a relatively recent version of Neovim rather than requiring the strict most recent. Or at least make this clear in the requirements since v2 supported older versions and v3 dropped all older neovim version support.

Thanks so much for your work on this project!

### Neovim version (nvim -v)

0.8.2

mehalter:
This should be tested, but I think this should fix with older Neovim compatibility.

I can test this fully early next week. Thanks to the maintainers for this great project!

Closes #658

### DIFF
--- a/lua/ibl/highlights.lua
+++ b/lua/ibl/highlights.lua
@@ -15,7 +15,14 @@ local M = {
 }
 
 local get = function(name)
-    return vim.api.nvim_get_hl(0, { name = name })
+    local hl
+    if vim.api.nvim_get_hl then -- check for new neovim 0.9 API
+        hl = vim.api.nvim_get_hl(0, { name = name })
+    else
+        ---@diagnostic disable-next-line
+        hl = vim.api.nvim_get_hl_by_name(name, true)
+    end
+    return hl
 end
 
 local not_set = function(hl)


lua/ibl/highlights.lua
```diff
@@ -15,7 +15,16 @@ local M = {
...
 local get = function(name)
-    return vim.api.nvim_get_hl(0, { name = name })
+    local hl
+    if vim.api.nvim_get_hl then -- check for new neovim 0.9 API
+        hl = vim.api.nvim_get_hl(0, { name = name })
+    else
+        hl = vim.api.nvim_get_hl_by_name(name, vim.o.termguicolors)
+        hl.fg, hl.bg = hl.foreground, hl.background
+        hl.ctermfg, hl.ctermbg = hl.fg, hl.bg
+        hl.sp = hl.special
```
lukas-reineke, line 25: this only covers the colors, but `nvim_get_hl` returns much more. It needs to cover all of these fields.

- fg (or foreground): color name or "#RRGGBB", see note.
- bg (or background): color name or "#RRGGBB", see note.
- sp (or special): color name or "#RRGGBB"
- blend: integer between 0 and 100
- bold: boolean
- standout: boolean
- underline: boolean
- undercurl: boolean
- underdouble: boolean
- underdotted: boolean
- underdashed: boolean
- strikethrough: boolean
- italic: boolean
- reverse: boolean
- nocombine: boolean
- link: name of another highlight group to link to, see |:hi-link|.
- default: Don't override existing definition |:hi-default|
- ctermfg: Sets foreground of cterm color |ctermfg|
- ctermbg: Sets background of cterm color |ctermbg|
- cterm: cterm attribute map, like |highlight-args|. If not set, cterm attributes will match those from the attribute map documented above.
- force: if true force update the highlight group when it exists.

lukas-reineke (changes requested):

lua/ibl/highlights.lua
mehalter: This function still carries all of those things. I removed the unnecessary renaming here.

mehalter:
@lukas-reineke just got around to testing this and it's working perfectly on neovim v0.8! Thanks for considering this PR 🙏🏼

lua/ibl/highlights.lua
```diff
@@ -15,7 +15,13 @@ local M = {
 }
 
 local get = function(name)
-    return vim.api.nvim_get_hl(0, { name = name })
+    local hl
+    if vim.api.nvim_get_hl then -- check for new neovim 0.9 API
+        hl = vim.api.nvim_get_hl(0, { name = name })
+    else
+        hl = vim.api.nvim_get_hl_by_name(name, true)
```
lukas-reineke, line 22: to make CI happy

```suggestion
        ---@diagnostic disable-next-line:deprecated
        hl = vim.api.nvim_get_hl_by_name(name, true)
```

lukas-reineke (changes requested):

lukas-reineke:
hm looks like the types for 0.9.2 just don't include it at all anymore. Just ignore all errors for that line I guess
`---@diagnostic disable-next-line`

lukas-reineke (approved):

Issue #649, opened 2023-09-28
Highlight current indent (no color) with a lower priority than scope (with color)


Danielkonge:
### Problem

I just updated to version 3 today and haven't yet played around with all the options, but I couldn't find something doing what I am asking for in the following. If it already exists, could someone let me know how to do it?

I really like the new highlighting via scope (especially in Rust). Even in Lua, I think this is helpful:
<img width="632" alt="Screenshot 2023-09-28 at 23 34 33" src="https://github.com/lukas-reineke/indent-blankline.nvim/assets/7075380/38534f20-a02f-4e9b-ab1a-958463218321">

So I plan to keep the highlighting on (I am matching it with rainbow-delimiters).

But whenever there is no scope highlighting I would like to fall back to slight highlighting of my current indentation level. I.e., in this picture:
<img width="627" alt="Screenshot 2023-09-28 at 23 34 56" src="https://github.com/lukas-reineke/indent-blankline.nvim/assets/7075380/626a1620-9a13-464c-b508-55454bc7c579">

I would like to have some slight highlighting of the second indent line to the left of my cursor. 

Preferably I would probably want to always have a slight highlighting of the current indent level unless that highlighting is overwritten by the scope highlighting. So sometimes the might be two highlighted indent lines at the same time. E.g., here:

<img width="538" alt="Screenshot 2023-09-28 at 23 46 15" src="https://github.com/lukas-reineke/indent-blankline.nvim/assets/7075380/ca708c53-04cc-4ab8-b0e2-ed11c238fffb">

I would like to keep the first indent red, but have the second indent slightly light up to indicate the current indent level at the cursor position. 

Is there already a way to do this? If not, I am willing to spend some time to try helping with implementation or testing of something like that. And if nothing else, I would like to request a feature like that. 

### Expected behavior

Always slightly highlight (e.g., just brighter color) the current indent level unless the current indent level is also the scope level in which case the scope is preferred for highlighting. 

(Alternatively, when not using the current scope to highlight anything, highlight slightly the current indent level.)

lukas-reineke:
No, it doesn't exist. There is a similar discussion in #632 to give some context.
In short, this would be difficult to implement and maintain to the level of quality I want. I don't think I will work on this for now.

Of course if someone makes a great PR I am open to add this. If anyone wants to try, please discuss the implementation with me first, so you don't waste time on a suboptimal solution.

lukas-reineke:
[...]

PR #668, opened 2023-09-30
VERY rough first draft of context in ibl.

Danielkonge:
This is a very rough first draft of an implementation of some kind of `context` in addition to `scope` in ibl. I don't have time to work more on this today, but please note that I am just uploading this to see if you already had any comments about changes you would want. This is not close to ready for a commit yet.

Right now the code is not cleaned up at all, and I will have to change a few things still. But I do have basically the behavior I want (see #649) now. 

For this first draft most of `context` is just all copied from `scope` and then I made a few minor changes.

### DIFF
--- a/after/plugin/commands.lua
+++ b/after/plugin/commands.lua
@@ -51,10 +51,6 @@ end, {
     desc = "Toggles indent-blanklines scope on and off",
 })
 
-
-
-
-
 vim.api.nvim_create_user_command("IBLEnableCurrentIndent", function()
     ibl.update { current_indent = { enabled = true } }
 end, {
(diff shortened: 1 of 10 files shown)


lukas-reineke:
You basically just implemented `scope = { include = { node_type = { ["*"] = { "*" } } } }`.
But like I mentioned here https://github.com/lukas-reineke/indent-blankline.nvim/issues/649#issuecomment-1740978705 this has too many edge cases. It's not a good enough solution to build a whole feature on.

Please think about the algorithm first. I don't want you to waste time implementing all the other stuff. Without a good algorithm, I am not going to merge this.

Danielkonge:
I will close this pull request for now and submit one without tree sitter for current indent highlighting. That one works on all my cases I tried, but it is slower.

PR #679, opened 2023-10-03
`current_indent` without treesitter

Danielkonge:
Here is an implementation of `current_indent` without using tree sitter. It works on all test cases I tried (though I haven't tried that many), but it is slower than the pure tree sitter stuff. Therefore I have tried to write it in such a way that if you disable it, it should have no noticeable impact on the performance of ibl. 

Would you be open to accepting a pull request similar to this?

(It needs to be cleaned up a bit more. E.g., if you use different `char` symbols, I haven't taken that into account yet. Also, there is some problem with the validation that doesn't work correctly when I edit my own config (it works as expected if I do stuff directly from the ibl files).)

### DIFF
--- a/lua/ibl/config.lua
+++ b/lua/ibl/config.lua
@@ -62,7 +62,7 @@ M.default_config = {
         },
     },
     current_indent = {
-        enabled = false,
+        enabled = true,
         char = nil,
         priority = 2,
         highlight = "IblCurrentIndent",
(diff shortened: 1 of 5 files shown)


lukas-reineke:
Please discuss what you want to built first! I feel bad to tell you no after you built it.
But sorry, but this is not good enough. Writing the extmarks, and then reading them again just to write them again, is way too slow. Performance is the first priority.
Even if it's not enabled by default. People will enable it and have a bad time, it's not acceptable.

Danielkonge:
Please don't feel sorry about me spending time on building this and you not accepting the pull request. I am mainly doing this because I want something like this for my own editing (so I can just use my own fork when I am satisfied with it) and to get more comfortable writing complex things in Lua (I am still pretty new to Lua). (Also I have time right now.)

Of course it would be nice to get to a level, where you would be happy to accept a pull request, but I am also fine with just maintaining a fork for myself.

PR #685, opened 2023-10-03
fix: ensure full refresh when color scheme changes

gaesa:
This commit fixes a bug in the highlight setup in `highlights.lua`. The bug was causing highlights not to refresh when the color scheme is changed. The fix involves introducing an optional `force_refresh` parameter to the `setup` and `setup_builtin_hl_groups` functions. When `force_refresh` is true, the highlights are refreshed even if they are already set.

Alternative Solution:
An alternative solution to this bug could involve extracting the three `vim.api.nvim_set_hl` calls into a new function (e.g., `force_refresh`). This function would be called in the callback of `autocmds.setup` function, ensuring that the highlights are always refreshed when this callback is executed. This approach could potentially make the code more modular and easier to understand, but it would also introduce more changes to the current codebase.

### DIFF
--- a/lua/ibl/autocmds.lua
+++ b/lua/ibl/autocmds.lua
@@ -52,7 +52,7 @@ M.setup = function()
         group = group,
         pattern = "*",
         callback = function()
-            highlights.setup()
+            highlights.setup(true)
             ibl.refresh_all()
         end,
     })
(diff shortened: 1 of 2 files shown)


lukas-reineke:
> The bug was causing highlights not to refresh when the color scheme is changed.

The color scheme needs to reset the highlight groups. If it doesn't, it's a bug in the color scheme, not indent-blankline.

gaesa:
I can confirm that this bug was not present in `v2.20.8`. However, I am not certain whether we should attribute the issue to the color scheme.

PR #703, opened 2023-10-05
Fast `current_indent` implementation

Danielkonge:
This is the most up to date implementation mentioned in #649, and I don't feel any slowdown compared to the version without `current_indent`.

Please let me know if you want me to change something, then I can edit it during the next few days.


Danielkonge:
Even if you end up not accepting this pull request, I think it is worth considering changing utils similar to what I have done with `start_pos_scope` and `end_pos_scope` in `utils.lua`. It worked more consistently for me at least. (Even without the old definitions too, I just left them to have more chances of hitting the color you want.)

lvim-tech:
indent
@Danielkonge thanks for your work. I think indentation doesn't work correctly in all cases

Danielkonge:
I found a bug where `current_indent` wasn't correct if you jumped back more than one indent level in a single line. I have fixed it now. I think that is what affected you too (going two levels back from line 370 to line 371). Could you let me know if the problem is fixed?

lvim-tech:
@Danielkonge now work correctly!

Danielkonge:
By now this commit has actually got a few different features mixed into it. 

It has `current_indent` as stated in the title of this pull request, but since this is really just my working implementation of indent-blankline, more stuff has gotten included. 

It now also adds a `show_exact_scope` and a `current_block_only` option + some improvements to the highlight hook for the scope. 

Would you prefer that I split these out into multiple smaller pull requests @lukas-reineke? Or is it easier for you to do it all at once in one bigger file (like now)?

lukas-reineke:
Please make smaller PRs :+1:

Danielkonge:
I finished the new smaller and cleaner pull requests now.

PR #704, opened 2023-10-05
Fix typo that prevents show_end from working as intended

Danielkonge:
Without this fix, the scope_end is not underlined unless you include a space or similar in front of the scope_end line. If the previous behavior is the intended behavior, I think there should be an option so you can get the behavior of this fix.

If we want to avoid highlighting a single parenthesis, we can just check that the length of the scope_end is greater than 1. 

Without fix:

With fix (ignore the white line, that is from the other pull request):

Screenshot 2023-10-06 at 01 05 15

### DIFF
--- a/lua/ibl/init.lua
+++ b/lua/ibl/init.lua
@@ -373,7 +373,7 @@ M.refresh = function(bufnr)
         end
 
         -- Scope end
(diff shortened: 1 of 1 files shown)


lukas-reineke:
This is on purpose.
Having the underline on the last line all the time is too noisy, and it also looks weird if the line is broken by the line.

PR #705, opened 2023-10-05
fix: extend indent lines through skipped lines

0xAdk:
indent-blankline-pr-example

Not sure if this should be a feature or a bug. I went with bug since it *feels* like the indent shouldn't break for skipped lines at 0 indent.

Not sure how this impacts performance, in a simple ad-hoc test it was slower by 1.88% but I didn't repeat it so that might just be variance. I feel like this should be faster since it moves the `hooks.get` call outside of the main loop, but idk.

### DIFF
--- a/lua/ibl/init.lua
+++ b/lua/ibl/init.lua
@@ -245,18 +245,34 @@ M.refresh = function(bufnr)
 
     local last_whitespace_tbl = {}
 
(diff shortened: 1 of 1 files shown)


lukas-reineke:
Yeah this is a bug, thanks
Please format the code with stylua, rest looks fine

PR #721, opened 2023-10-10
ci: add conventional commit message validation

lukas-reineke:


lukas-reineke:
why do all prebuilt actions for this don't work :cry:

PR #723, opened 2023-10-10
Only show indent guides for current block (resolves #561)

### LINKED ISSUE #561: Feature Request: Only show indent lines on my current code "block"
Say I have this code block
```
impl Location {
│ pub async fn get_by_id(id: i32) -> Result<Self, Error> {
│ │ Ok(sqlx::query_as!(Self, r#"SELECT * FROM location WHERE location_id = $1"#, id).fetch_one(&ctx.pool).await?)
│ }
}

impl Address {
│ pub async fn get_by_id(id: i32) -> Result<Self, Error> {
│ │ Ok(sqlx::query_as!(Self, r#"SELECT * FROM location WHERE location_id = $1"#, id).fetch_one(&ctx.pool).await?)
│ }█
}
```
Notice the cursor is in the second code block. I would hope the guides would only show under the `impl Address` block:

```
impl Location {
  pub async fn get_by_id(id: i32) -> Result<Self, Error> {
    Ok(sqlx::query_as!(Self, r#"SELECT * FROM location WHERE location_id = $1"#, id).fetch_one(&ctx.pool).await?)
  }
}

impl Address {
│ pub async fn get_by_id(id: i32) -> Result<Self, Error> {
│ │ Ok(sqlx::query_as!(Self, r#"SELECT * FROM location WHERE location_id = $1"#, id).fetch_one(&ctx.pool).await?)
│ }█
}
```
I think it could look cleaner

Danielkonge:
This should resolve #561. If you want anything to change I can edit it later today. 

One thing to consider is if there should be an option to change the behavior when you are not in any code block. Right now it will not show any indent guides when you are not in a code block, and then only show the blocks indent guides when you are in a code block. I think this could be useful for viewing large files with multiple different blocks, which is why I picked this behavior as the standard. 

Also, I am not sure exactly how you would want to write tests for this, but if you point me in the right direction, I can also try to look into that.


lukas-reineke:
I think it makes more sense if this builds on current indent right?

Danielkonge:
You are right, but I thought some people will want to use this without using `current_indent`. But I guess I can just make it so that you only do the necessary calculation if either `current_indent` needs it or this needs it. I will make a new pull request with those features both then.

Danielkonge:
Sorry, I will reopen this, but I improved it so it doesn't use scope now. It was not a good idea to use scope since it doesn't capture the right thing, but I do think we want to use treesitter for this. Without treesitter, it can be hard to describe the exact block you are in (this is a different thing than what `current_indent` does, since it doesn't care where the block starts/ends, it just looks at where the indent guides start/end). 

All languages I have tested (and I think this should generally be true for treesitter?) has new code blocks as the child of the root, so going up the tree until you see the root should always work (if my assumption is true).

lukas-reineke:
I don't think treesitter is the right tool for this.  
This is basically current indent, but walking up all indent levels until you hit the root.

For current indent, you keep a stack of all indents until you hit the cursor, then display the deepest one differently.  
For this feature, you would display all indents in the stack, and nothing else.

Danielkonge:
Ah, I think I misunderstood what you said originally. Keeping a stack similar to current_ident but for current_block like what you described should definitely be possible. I can try to do that later. 

I think I will wait until `current_indent` has been accepted though, so I will know what I can use from that pull request.

PR #724, opened 2023-10-10
Added a show exact scope option

Danielkonge:
This adds an option that lets you clearly see the exact start and end of the scope. It might not always look as nice as the default setup, so it is disabled by default, but I have found it useful and I use this in my own setup now.


lua/ibl/init.lua
```diff
@@ -390,9 +391,18 @@ M.refresh = function(bufnr)
         -- #### set virtual text ####
         vt.clear_buffer(bufnr, row)
 
+        -- Show exact scope
+        local scope_col_start_draw = #whitespace
+        local show_end_cond = #whitespace_tbl > scope_col_start_single
```
lukas-reineke, line 396:
```suggestion
        local scope_show_end_cond = #whitespace_tbl > scope_col_start_single
```

lukas-reineke (changes requested):
I still think it looks pretty broken, but I'm okay with adding it.

Just a small comment to make the variable name more clear

lukas-reineke (approved):

PR #726, opened 2023-10-12
Fast current_indent (resolves #649)

### LINKED ISSUE #649: Highlight current indent (no color) with a lower priority than scope (with color)

Danielkonge:
Here is a cleaned up version of my implementation of `current_indent` that doesn't contain other features.

This pull request will resolve #649.

### DIFF
--- a/lua/ibl/init.lua
+++ b/lua/ibl/init.lua
@@ -263,7 +263,15 @@ M.refresh = function(bufnr)
         end
     end
 
-    -- a table
+    ---@type number[]
+    local next_drawn_indents
+
+    ---@type number
+    local cur_indent_stack_size = 0
+
+    ---@type number
+    local current_indent_col = -1
+
     ---@type ibl.indent.whitespace[]
     local next_whitespace_tbl = {}
 
@@ -314,10 +322,8 @@ M.refresh = function(bufnr)
         ---@type ibl.indent.whitespace[]
         local whitespace_tbl
 
-        local prev_indent_stack_size = 0
-        if config.current_indent.enabled and indent_state then
-            prev_indent_stack_size = #indent_state.stack or 0
-        end
+        ---@type number[]
+        local drawn_indents
 
         local blankline = line:len() == 0
         arr_blankline[i] = blankline
@@ -327,10 +333,11 @@ M.refresh = function(bufnr)
 
         -- #### calculate indent ####
         if not blankline then
-            whitespace_tbl, indent_state = indent.get(whitespace, indent_opts, indent_state)
+            whitespace_tbl, indent_state, drawn_indents = indent.get(whitespace, indent_opts, indent_state)
         elseif empty_line_counter > 0 then
             empty_line_counter = empty_line_counter - 1
             whitespace_tbl = next_whitespace_tbl
+            drawn_indents = next_drawn_indents
         else
             if i == #lines then
                 whitespace_tbl = {}
@@ -342,7 +349,7 @@ M.refresh = function(bufnr)
                 end
 
                 local j_whitespace = utils.get_whitespace(lines[j])
-                whitespace_tbl, indent_state = indent.get(j_whitespace, indent_opts, indent_state)
+                whitespace_tbl, indent_state, drawn_indents = indent.get(j_whitespace, indent_opts, indent_state)
 
                 if utils.has_end(lines[j]) then
                     local last_whitespace_tbl = arr_whitespace_tbl[i - 1] or {}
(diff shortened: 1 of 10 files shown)


Danielkonge:
@lukas-reineke I think this should be basically ready now, but I cannot figure out what stylua is complaining about? I think it might just not like the `::continue::` syntax, but it weirdly accepts the later one. Can you tell what is wrong?

lua/ibl/init.lua
```diff
@@ -365,13 +418,78 @@ M.refresh = function(bufnr)
...
-        last_whitespace_tbl = whitespace_tbl
+        -- this do block is purely to help stylua with the scope
+        do
+            arr_whitespace_tbl[i] = whitespace_tbl
+        end
+
+        ::continue::
+    end
+
+    -- between the two loops we find the current indent column and current indent start quickly
```
lukas-reineke, line 429: Why does this need to happen between the loop? If you keep track of the current indent in the first loop, you can just use that as the current indent when you hit the current line

lukas-reineke (changes requested):
Current indent should also support `show_start` and `show_end`.
And you need to add `priority` that decides if scope or current indent is displayed if they are in the same place.

lukas-reineke:
I get this behavior with tabs. The second if block should not be highlighted

lua/ibl/init.lua
Danielkonge: I have updated this now so it happens in the loop.

Danielkonge:
> I get this behavior with tabs. The second if block should not be highlighted
> 
> 

This should be fixed now. (Please let me know if it not.) I used `indent_state.stack` to decide what indents were drawn, but I can see it doesn't work as I would expect in all cases (e.g. with the blank lines in the if statements like in the picture). To overcome this I have constructed a table `drawn_indents` alongside the `whitespace_tbl` that keeps track of exactly which indents have been drawn. This also simplifies a workaround with blank lines I used earlier.

Danielkonge:
I have added `priority`, `show_start` and `show_end` now. Note that `show_end` doesn't really make sense in the same way for the indentation guides as for scope, so the only thing I think makes sense here is to underline the line below the end of the current indent level if you enable it.

lukas-reineke:
I don't quite understand the need for `drawn_indents`, isn't it just the same information you already have in `whitespace_tbl`?

Danielkonge:
> I don't quite understand the need for `drawn_indents`, isn't it just the same information you already have in `whitespace_tbl`?

The information in `drawn_indents` is already in `whitespace_tbl`, yes. I just thought it was easier to work with a list of purely the indents, whereas `whitespace_tbl` includes spaces and other stuff too. If you prefer, I can rewrite it to just use `whitespace_tbl` directly (you will just need to search backwards through `whitespace_tbl` for the indents).

`drawn_indents` just contains the positions of `TAB_START`, `TAB_START_SINGLE` and `INDENT` in the `whitespace_tbl` (and it ignores `TAB_FILL`, `TAB_END` and `SPACE`), since these are the ones used for drawing the indentation guides.

lukas-reineke:
👍 yeah that's what I mean. Let's just use `whitespace_tbl`. You can make a helper function to give you back the indents.
It shouldn't hurt performance, and it's easier to understand when there is only one data source.

Danielkonge:
> 👍 yeah that's what I mean. Let's just use `whitespace_tbl`. You can make a helper function to give you back the indents. It shouldn't hurt performance, and it's easier to understand when there is only one data source.

Since I only actually really need the last indent column and the number of indents I have made a helper function to return these given a whitespace table now.

Danielkonge:
Since the other pull request is almost ready, I will close this one.

wookayin:
Continued in #743 (adding a ref here because they were not linked). Thanks @Danielkonge for the great work.

Issue #778, opened 2023-11-30
Add better control of inclusion/exclusion of certain file/buf/...-types 


Danielkonge:
### Problem

I would like to use `indent-blankline` in `InspectTree`, but `InspectTree` is of buftype `'nofile'`. Of course I could just remove `'nofile'` from the excluded buftypes, but I do actually want to exclude `'nofile'` in all other cases. Therefore it would be nice to have slightly better control of whether to use `indent-blankline` in certain cases. 

### Expected behavior

I am not sure what the best solution to this would be with the way `indent-blankline` works right now. 

In `rainbow-delimiters` I have done something slightly similar in highlighting of `query` files, where I highlight one way if `vim.b.dev_base` is truthy (i.e., we are in `InspectTree`) and another way if `vim.b.dev_base` is `nil`. I can do this here by allowing setup via a function that I then use the return value of. 

If you have an idea for how to do something like I want, I can implement a draft of it, if you would want that. 

Alternatively, is there already a way to do what I want that I missed?

Danielkonge:
Update: I tried removing `'nofile'` from the exluded buftypes and adding the following hook:
```lua
hooks.register(
    hooks.type.ACTIVE,
    function(_)
        if vim.bo.buftype == 'nofile' then
            if vim.b.dev_base then
                return true
            else
                return false
            end
        end
        return true
    end
)
```
But this didn't work. However when I go and edit out `'nofile'` from the exclude list in the source code, then it works, so it seems there is still some problem with updating the config correctly. I will try to find the problem.

lukas-reineke:
> Add a way to overwrite the excludes.

There is, it's called `:help ibl.overwrite()`. With this, you can completely replace excludes instead of adding to it. Then your hook should work fine.

Danielkonge:
`ibl.overwrite()` didn't give me the behavior I want. I think it is because of this line:
https://github.com/lukas-reineke/indent-blankline.nvim/blob/29be0919b91fb59eca9e90690d76014233392bef/lua/ibl/config.lua#L274

I changed
```lua
    M.config = merge_configs("merge", M.default_config, config)
```
into
```lua
    M.config = merge_configs("merge", M.config or M.default_config, config)
```
and now I get the expected behavior of `ibl.overwrite()` using 
```lua
ibl.overwrite({
    exclude = {
        buftypes = {
            "terminal",
            -- don't include "nofile",
            "quickfix",
            "prompt",
        }
    }
})
[...]

PR #779, opened 2023-12-01
fix: set_config

Danielkonge:
This code fixes the problem I mention with `ibl.overwrite()` in #778. With this bug fix I get the expected behavior mentioned in that issue. (This also closes that issue, since I get the behavior I was asking for.)

### DIFF
--- a/lua/ibl/config.lua
+++ b/lua/ibl/config.lua
@@ -271,7 +271,7 @@ end
 ---@return ibl.config.full
 M.set_config = function(config)
     validate_config(config)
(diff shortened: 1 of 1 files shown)


lukas-reineke:
This is wrong, `setup` and by extension `set_config` explicitly resets to the default. From the help

>  `setup` is idempotent, meaning you can call it multiple times, and each call will reset indent-blankline. If you want to only update the current configuration, use |ibl.update()|.

Also, `ibl.overwrite` does not call `set_config`, I'm not sure I understand what your problem is.

PR #780, opened 2023-12-01
docs: note that overwrite should be after setup

Danielkonge:
I think adding this to the docs will help clarify the use of `ibl.overwrite()`.

### DIFF
--- a/doc/indent_blankline.txt
+++ b/doc/indent_blankline.txt
@@ -36,7 +36,7 @@ setup({config})                                                  *ibl.setup()*
 
  `setup` is idempotent, meaning you can call it multiple times, and each call
  will reset indent-blankline. If you want to only update the current
- configuration, use |ibl.update()|.
+ configuration, use |ibl.update()| or |ibl.overwrite()|.
 
  Parameters: ~
    • {config}  (|ibl.config|?) Configuration table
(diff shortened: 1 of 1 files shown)


lukas-reineke:
Hm this isn't really correct either. You can use `overwrite` without `setup`. The important point is that `setup` resets everything, and that is already covered in the documentation.

Maybe the documentation for `setup` can include overwrite.

```diff
`setup` is idempotent, meaning you can call it multiple times, and each call
will reset indent-blankline. If you want to only update the current
- configuration, use |ibl.update()|.
+ configuration, use |ibl.update()| or |ibl.overwrite()|.
```

Danielkonge:
> Hm this isn't really correct either. You can use `overwrite` without `setup`. The important point is that `setup` resets everything, and that is already covered in the documentation.
> 
> Maybe the documentation for `setup` can include overwrite.
> 
> ```diff
> `setup` is idempotent, meaning you can call it multiple times, and each call
> will reset indent-blankline. If you want to only update the current
> - configuration, use |ibl.update()|.
> + configuration, use |ibl.update()| or |ibl.overwrite()|.
> ```

That would be a good addition too, but I still think it makes sense to include something under `ibl.overwrite()` too. What about something like:
```diff
-  Note that |ibl.overwrite| should be called after |ibl.setup|.
+  Note: Since |ibl.setup()| resets indent-blankline, you will need to call 
+  |ibl.overwrite()| after |ibl.setup()| to avoid loosing your overwrites.
```
Would that be better? 

I assume that most users of the plugin use `|ibl.setup()|`, so I think it makes sense to mention this under `|ibl.overwrite()|` too.

lukas-reineke:
I want to avoid making it sound like you have to use `setup`.

Maybe something like `In case you use both |ibl.setup()|, and |ibl.overwrite()|, make sure to call setup first`

lukas-reineke (approved):

lukas-reineke:
LGTM, thanks

Issue #776, opened 2023-11-24
Performance drop in huge files (over 1200 LOC)


hinell:
### Problem

Slow in files with over 1200+ LOCs.

Here is a log of called functions obtained by `jit.p`. Numbers signify times the given function is called:
```lua
autocmds.lua:25;debounced_refresh;debounced_refresh;refresh;get 504
highlighter.lua:292 461
autocmds.lua:25;debounced_refresh;debounced_refresh;refresh 432
highlighter.lua:292;on_line_impl;for_each_tree;fn;iter 362
autocmds.lua:25;debounced_refresh;debounced_refresh;refresh;get;is_indent 356
highlighter.lua:292;on_line_impl;for_each_tree;fn;iter;match_preds;handler 330
autocmds.lua:25;debounced_refresh;debounced_refresh;refresh;get;tbl_filter 307
highlighter.lua:292;on_line_impl;for_each_tree;fn 302
autocmds.lua:39;debounced_refresh;debounced_refresh;refresh;get 268
autocmds.lua:39;debounced_refresh;debounced_refresh;refresh 215
autocmds.lua:39;debounced_refresh;debounced_refresh;refresh;get;is_indent 215
autocmds.lua:25;debounced_refresh;debounced_refresh;refresh;get;tbl_contains 213
treesitter-context.lua:92;get_context;get_context_query;pcall;_memoize.lua:51;fn;fn 198
init.lua:19 193
```

`debounced_refresh()` is called too often.  
The issue may be partially mitigated by increasing `config.debounce = 768` but it isn't perfect.

### Steps to reproduce

1. Config:
```lua
-- $HOME/.config/nvim/lua/user/plugins/indent-blankline.lua
local config = require("ibl.config").default_config
config.debounce        = 200

require("ibl").setup(config)
```

3. Open the following file:
```bash
#!/usr/bin/bash
nvim -c "set ft=c" <(curl -L https://github.com/tree-sitter/tree-sitter-bash/raw/master/src/scanner.c)
```

### Expected behavior

Slick & fast

### Neovim version (nvim -v)

NVIM v0.10.0-dev-8405649f9

lukas-reineke:
The only thing that gets slower with bigger files is scope, because of treesitter. You can turn it off in big files, or use debounce to reduce the number of times indent-blankline refreshes. But there isn't really anything else I can do.

qRoC:
[...]

PR #790, opened 2023-12-06
perf: prevent multiple calls of M.refresh() by autocmds

hinell:
This PR:
* eliminates performance bottleneck in big files
* prevents multiple async calls to `M.refresh()`, by autocmds
* refactor debounce timer

### Overview
While profiling, [I've discovered](https://github.com/lukas-reineke/indent-blankline.nvim/issues/776#issuecomment-1842205823) that `M.refresh()` is actually called twice many times. This should be fixed now.

Demo (run in bash) 50K LOCs:
```bash
cd PATH_TO_IBL_REPO
gh pr checkout 790
nvim -c "set ft=typescript" <(curl -L https://raw.githubusercontent.com/microsoft/TypeScript/main/src/compiler/parser.ts)
```

### Related
* #776


lukas-reineke:
There are probably cases where this can be optimized, but this PR removes important functionality.

The first call to `debounced_refresh` should not be debounced, or there is noticeable lag when first opening a buffer, or jumping to a section that was not rendered yet.
Creating and removing the timer on each refresh is going to hurt performance.
There are cases that need to take priority and rerender, just not allowing this will cause bugs.

lukas-reineke:
> There are cases that need to take priority and rerender, just not allowing this will cause bugs.

This is still true. Just not allowing the refresh is not possible.
I took another look at the code, and there is one small issue I found that leads to multiple refresh calls. But when testing it did not happen often enough to make a difference I think. #791

hinell:
> > There are cases that need to take priority and rerender, just not allowing this will cause bugs.
> 
> This is still true. Just not allowing the refresh is not possible.

This PR skips `M.refresh` from running again if it's already running asynchronously by some other autocmd. Lua puts multiple `M.refresh` calls for the same buffer on stack producing huge overhead otherwise.

Btw, the lag is still noticiable even with old timer approach.

lukas-reineke:
> This PR skips M.refresh from running again if it's already running asynchronously by some other autocmd.

Which will cause bugs. Some events need to run even if the buffer is already being refreshed, like changes to the config. Every none important event goes through the debounce already.

> Lua puts multiple M.refresh calls for the same buffer on stack producing huge overhead otherwise

If this happens to you, it probably means the execution of refresh takes longer than the debounce. In that case, you need to set debounce higher. I am fairly confident that there are no duplicate triggers of refresh that don't get debounced. If there are, please show me where.

hinell:
The performance gain is obvious. I haven't observed any bugs so far.

I wouldn't really want `M.refresh()` to race, regardless what it does inside.

Best.

lukas-reineke:
> The performance gain is obvious

You will get the same performance gain by increasing debounce. This is working like it should.
You can easily test it by adding a counter to refresh, and setting debounce to a high value, like 1 second. Refresh won't be called more than once a second.

> I wouldn't really want M.refresh() to race, regardless what it does inside.

refresh updates the buffer linearly from top to bottom. It is safe, and necessary, to run it before the previous execution is done. *But* only in special cases, normal moving around the buffer will not cause this.

Unless refresh takes longer than what debounce is set to (in which case you need to update your config), there won't be multiple calls to refresh for the same buffer at the same time. With the exception of special events like config changes.

PR #795, opened 2023-12-15
fix: move some scope calculations out of main loop

### LINKED ISSUE #752: Bad indentation of scope for multi-line function definition
### Problem

I noticed an indentation issue of the scope highlighting when viewing C++ files. However I saw the same issue also in Java files so I assume it's a general issue.

My project puts the opening brace `{` in the same line as the if-clause, function parameters, etc. If the parameter list spans multiple lines then it uses the wrong indentation to display the scope (see images).

image

image

### Steps to reproduce

* This is using indent-blankline v3.
* Scopes must be enabled.
* Create a C++ file and a function with a multi-line parameter list as shown in the images.
* Move the cursor inside the function

### Expected behavior

It should instead highlight the leftmost line in case of the function scope and in case of the if-clause, it should highlight the 2nd line, aligning with the indentation of the `if`.

### Neovim version (nvim -v)

0.9.1

Danielkonge:
This pull request moves the calculation of the scope index out of the main loop, since we want to be able to figure this out without including lines all the way from the start of the scope if we are in the middle of a very long scope. 

Since we no longer need to include as many lines to get good highlighting, I also lowered the default value of `viewport_buffer.max` from 500 to 60. I think we can lower both `min` and `max` even more, but at that point it becomes a question of personal preference. I picked 60, since that should about cover one full screen for most people (I think?). 

Furthermore, I added a comparison between the start line and end line whitespace tables, and we use the smaller one to calculate the scope index. This fixes #752, and I can't think of any normal formatting style that will get worse highlighting by this. If you want, I can include an option for whether to use the start (of the scope), end or the minimum of both?

### DIFF
--- a/lua/ibl/init.lua
+++ b/lua/ibl/init.lua
@@ -193,9 +193,8 @@ M.refresh = function(bufnr)
         if scope and scope:start() >= 0 then
             local scope_start = scope:start()
             local scope_end = scope:end_()
-            offset = top_offset - math.min(top_offset - math.min(offset, scope_start), config.viewport_buffer.max)
-            scope_start_line = vim.api.nvim_buf_get_lines(bufnr, scope_start, scope_start + 1, false)
-            scope_end_line = vim.api.nvim_buf_get_lines(bufnr, scope_end, scope_end + 1, false)
+            scope_start_line = vim.api.nvim_buf_get_lines(bufnr, scope_start, scope_start + 1, false)[1]
+            scope_end_line = vim.api.nvim_buf_get_lines(bufnr, scope_end, scope_end + 1, false)[1]
         end
     end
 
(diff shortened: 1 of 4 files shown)


lua/ibl/init.lua
```diff
@@ -187,10 +187,15 @@ M.refresh = function(bufnr)
...
+    local scope_start_line, scope_end_line
     if not scope_disabled and config.scope.enabled then
         scope = scp.get(bufnr, config)
         if scope and scope:start() >= 0 then
-            offset = top_offset - math.min(top_offset - math.min(offset, scope:start()), config.viewport_buffer.max)
+            local scope_start = scope:start()
+            local scope_end = scope:end_()
+            offset = top_offset - math.min(top_offset - math.min(offset, scope_start), config.viewport_buffer.max)
+            scope_start_line = vim.api.nvim_buf_get_lines(bufnr, scope_start, scope_start + 1, false)
+            scope_end_line = vim.api.nvim_buf_get_lines(bufnr, scope_end, scope_end + 1, false)
```
lukas-reineke, line 198:
```suggestion
            scope_start_line = vim.api.nvim_buf_get_lines(bufnr, scope_start, scope_start + 1, false)[1]
            scope_end_line = vim.api.nvim_buf_get_lines(bufnr, scope_end, scope_end + 1, false)[1]
```

It's cleaner to do this here.

lua/ibl/init.lua
```diff
@@ -260,6 +265,49 @@ M.refresh = function(bufnr)
...
+            whitespace_tbl = whitespace_tbl_start
+        end
+
+        -- do the same calculations as in the main loop below, but note that the scope
+        -- start and end will never be on a blankline, so these cases simplify a lot
+        local current_left_offset = left_offset
+        while #whitespace_tbl > 0 and current_left_offset > 0 do
+            table.remove(whitespace_tbl, 1)
+            current_left_offset = current_left_offset - 1
+        end
```
lukas-reineke, line 290: The horizontal scroll fix is exactly the same as in the main loop, I think it would be nicer to make this a util function.

lua/ibl/init.lua
```diff
@@ -260,6 +265,49 @@ M.refresh = function(bufnr)
...
+
+        for _, fn in
+            pairs(hooks.get(bufnr, hooks.type.WHITESPACE) --[=[@as ibl.hooks.cb.whitespace[]]=])
+        do
+            whitespace_tbl = fn(buffer_state.tick, bufnr, scope_row_start - 1, whitespace_tbl)
```
lukas-reineke, line 295: > `scope_row_start`

This could be the end row instead. Depends on the condition from line 276

lua/ibl/init.lua
```diff
@@ -187,10 +187,15 @@ M.refresh = function(bufnr)
...
 
     local scope
+    local scope_start_line, scope_end_line
     if not scope_disabled and config.scope.enabled then
         scope = scp.get(bufnr, config)
         if scope and scope:start() >= 0 then
-            offset = top_offset - math.min(top_offset - math.min(offset, scope:start()), config.viewport_buffer.max)
+            local scope_start = scope:start()
+            local scope_end = scope:end_()
+            offset = top_offset - math.min(top_offset - math.min(offset, scope_start), config.viewport_buffer.max)
```
lukas-reineke, line 196: The only reason I adjust offset here at all was to compensate for the scope calculation. This is no longer needed with this PR. You can just remove this line.

And maybe add a note in the docs that `viewport_buffer.max` is deprecated. But don't remove it, we don't want to make this a breaking change.

lukas-reineke (changes requested):
A couple small points, but looks good overall.

lua/ibl/init.lua
Danielkonge: I fixed this.

lua/ibl/init.lua
Danielkonge: I made a util function now. Please let me know if you want me to change anything.

lua/ibl/init.lua
Danielkonge: I changed it so it matches end row when it that fits better.

lua/ibl/init.lua
Danielkonge: I made a small update to the docs now. I think it would also make sense to add small change to the `min` docs too, noting that it is now the exact offset and not minimal. Should I just add that too?

Danielkonge:
I updated this based on your comments now. If you have anything else you would like to update, I can do that too, and then I plan to squish the commits afterwards.

lukas-reineke (approved):
thanks :rocket:

PR #798, opened 2023-12-15
feat: add `CursorMovedI` autocmd

Danielkonge:
I don't move that much in insert mode usually, but when I do, it is nicer with this addition to the auto commands.


lukas-reineke (approved):
Thanks

PR #800, opened 2023-12-16
Added language: risor

applejag:
Adds language `risor`: <https://risor.io/>

I'm currently creating a tree-sitter grammar for it: <https://github.com/applejag/tree-sitter-risor>

Took me a while of debugging why the scope feature didn't work out-of-the box. I thought for so long that my grammar was wrong 😅 never considered that it was ibl

I was not able to add the language just from configs though. Would've been nice if that was possible. But currently ibl just skips if it does not contian the language in its hardcoded list:

https://github.com/lukas-reineke/indent-blankline.nvim/blob/d4c718467d35bc93714425a7102d82e7e5065280/lua/ibl/scope.lua#L60-L63

### DIFF
--- a/lua/ibl/scope_languages.lua
+++ b/lua/ibl/scope_languages.lua
@@ -416,6 +416,13 @@ local M = {
     re2c = {
         body = true,
     },
+    risor = {
+        function_declaration = true,
+        if_statement = true,
+        block = true,
+        switch_statement = true,
+        for_statement = true,
+    },
     ron = {
         array = true,
         map = true,


lukas-reineke (changes requested):
Thanks
Code looks fine, but can you use [conventional commit messages](https://www.conventionalcommits.org/en/v1.0.0/), please?

> I was not able to add the language just from configs though. Would've been nice if that was possible. But currently ibl just skips if it does not contian the language in its hardcoded list:

I thought about adding an option for this, but decided against it. It should be rare that someone adds a whole new language. And in case they do, I want to encourage them to make a PR, like you are now. Then it will work out of the box for everyone else.

applejag:
> And in case they do, I want to encourage them to make a PR, like you are now.

Ohohoh you sneaky fella. It sure worked!

lukas-reineke (approved):

PR #802, opened 2023-12-19
perf: make the scope from extmarks hook fast

griwes:
The current implementation of the hook to extract highlight info from extmarks is _really slow_. Slow to the point where, in neovide, there's a 1-2s lag between me starting to scroll with my mouse and anything happening if `current_scope` is enabled. The changes in this PR make the scrolling happen immediately, and I also no longer get any lags when moving my cursor by holding a down or up key, which is happening (especially with files larger than ~200ish lines) on master.

For the most part, this should be equivalent to the old code; in _theory_, there could be cases where the nested iteration yields a different answer, but I believe that it would require _two_ of the highlights we are looking for to be applied at the same buffer position, and I don't think that that is really an edge case worth preserving (but am happy to modify these changes if you'd prefer that we retain the exact same behavior there).

At first I thought it was just the excessive looping in the second part of the function that made it slow, but changing that has given me just moderate improvements; switching from `vim.inspect_pos` to `nvim_buf_get_extmarks` gave me a much higher speedup.

### DIFF
--- a/lua/ibl/hooks.lua
+++ b/lua/ibl/hooks.lua
@@ -200,6 +200,11 @@ M.builtin = {
             return scope_index
         end
 
+        local reverse_hls = {}
+        for i, hl in ipairs(highlight) do
+            reverse_hls[hl] = i
+        end
+
         local start_row, start_col = scope:start()
         local end_row, end_col = scope:end_()
         local start_line = vim.api.nvim_buf_get_lines(bufnr, start_row, start_row + 1, false)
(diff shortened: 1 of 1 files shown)


lukas-reineke (approved):
Thanks, looks good

> At first I thought it was just the excessive looping in the second part of the function that made it slow, but changing that has given me just moderate improvements

The loops only have a very low number of iterations. Like you say, it shouldn't really make a big difference.

> switching from vim.inspect_pos to nvim_buf_get_extmarks gave me a much higher speedup.

I wonder why it's so much faster. I read the implementation of `inspect_pos` when I wrote this, and it's basically just using `nvim_buf_get_extmarks` with some cleanup..
Maybe it's the calls to `vim.validate` again :thinking:

PR #805, opened 2023-12-21
feat: improve scope detection and calculation

### LINKED ISSUE #799: [Question] Python scopes are detected in the first line (header) for top-level scopes but not nested scopes.
This isn't an issue per se; I'm asking out of curiosity more than anything else. 

The behavior can be demonstrated using the following sample code (refer to the attached video):
```python
def top_level_function(): # For top-level blocks, the scope is highlighted when the cursor is at the header.

    print("foo")


class TopLevelClass:  # For top-level blocks, the scope is highlighted when the cursor is at the header.

    def __init__(self) -> None: # Which isn't the case for nested blocks.

        # For nested blocks, the scope is only highlighted when the cursor is one line past the header
        self.foo = "bar"
```
 Note that both `class_definition` and `function_definition` are listed in `scope.include.python`

https://github.com/lukas-reineke/indent-blankline.nvim/assets/69449791/d18c2175-7a77-4a64-bd42-7e38ba1a405b

Which seems a bit strange to me because upon inspecting the highlight groups using `:InspectTree`, I didn't see any difference in the way the two `function_definition`s are parsed other than the fact that one is nested in a block while the other is at the top level. 

```
(function_definition) ; [3:1 - 5:16]      <<<< This is the top-level function
 name: (identifier) ; [3:5 - 22]
 parameters: (parameters) ; [3:23 - 24]
 (comment) ; [3:27 - 108]
 body: (block) ; [5:5 - 16]
  (expression_statement) ; [5:5 - 16]
   (call) ; [5:5 - 16]
    function: (identifier) ; [5:5 - 9]
    arguments: (argument_list) ; [5:10 - 16]
     (string) ; [5:11 - 15]
      (string_start) ; [5:11 - 11]
      (string_content) ; [5:12 - 14]
      (string_end) ; [5:15 - 15]
(class_definition) ; [8:1 - 13:24]
 name: (identifier) ; [8:7 - 19]
 (comment) ; [8:23 - 104]
 body: (block) ; [10:5 - 13:24]
  (function_definition) ; [10:5 - 13:24]     <<<< This is the nested function
   name: (identifier) ; [10:9 - 16]
   parameters: (parameters) ; [10:17 - 22]
    (identifier) ; [10:18 - 21]
   return_type: (type) ; [10:27 - 30]
    (none) ; [10:27 - 30]
   (comment) ; [10:33 - 73]
   (comment) ; [12:9 - 102]
   body: (block) ; [13:9 - 24]
    (expression_statement) ; [13:9 - 24]
     (assignment) ; [13:9 - 24]
      left: (attribute) ; [13:9 - 16]
       object: (identifier) ; [13:9 - 12]
       attribute: (identifier) ; [13:14 - 16]
      right: (string) ; [13:20 - 24]
       (string_start) ; [13:20 - 20]
       (string_content) ; [13:21 - 23]
       (string_end) ; [13:24 - 24]
```
As you can see, in both `function_definition`s, the start and end positions listed by the parser are accurate. Given their similarities, I was wondering why the plugin treats both cases differently. According to the parser, the nested function begins at position `[10:5]`; but since the indent line is shown starting from one line below the declaration, the highlighting of the indent line will not take place until the cursor has gone past the first newline in the scope. This behavior is sensible albeit inconsistent with the case where the declaration is at the top-level of the source file. For `top_level_function()`, the highlighting takes place as soon as the cursor enters the line containing the declaration.

Danielkonge:
This uses the exact cursor position to find the treesitter node used when finding the scope, which improves scope detection in Python:

Screenshot 2023-12-21 at 17 47 25

Screenshot 2023-12-21 at 17 47 40

Importantly, it doesn't change the `language_for_range` calculation, which from what I can tell is the reason the other range was starting at column 0 of the current row. 

I also added an early exit of the scope calculation if the node didn't change (`TSNode:equal(TSNode)` tests that the tree and id are equal), so we can skip the slightly slower calculations after that step (joining and going through tables). This should slightly speed up the calculation in a lot of cases (given all the auto commands used), but it is less important than the other change.

This should close #799.

### DIFF
--- a/lua/ibl/init.lua
+++ b/lua/ibl/init.lua
@@ -189,7 +189,7 @@ M.refresh = function(bufnr)
     local scope
     local scope_start_line, scope_end_line
     if not scope_disabled and config.scope.enabled then
-        scope = scp.get(bufnr, config)
+        scope = scp.get(bufnr, config, global_buffer_state[bufnr] or {})
         if scope and scope:start() >= 0 then
             local scope_start = scope:start()
             local scope_end = scope:end_()
--- a/lua/ibl/scope.lua
+++ b/lua/ibl/scope.lua
@@ -4,10 +4,11 @@ local M = {}
 
 ---@param win number
 ---@return table<number, number>
-M.get_cursor_range = function(win)
+---@return table<number, number>
+M.get_cursor_ranges = function(win)
     local pos = vim.api.nvim_win_get_cursor(win)
     local row, col = pos[1] - 1, pos[2]
-    return { row, 0, row, col }
+    return { row, 0, row, col }, { row, col, row, col }
 end
 
 --- Takes a language tree and a range, and returns the child language tree for that range
@@ -34,8 +35,9 @@ end
 
 ---@param bufnr number
 ---@param config ibl.config.full
+---@param buffer_state { scope: TSNode?, left_offset: number, top_offset: number, tick: number }
 ---@return TSNode?
-M.get = function(bufnr, config)
+M.get = function(bufnr, config, buffer_state)
     local lang_tree_ok, lang_tree = pcall(vim.treesitter.get_parser, bufnr)
     if not lang_tree_ok or not lang_tree then
         return nil
@@ -51,7 +53,7 @@ M.get = function(bufnr, config)
         win = 0
     end
 
-    local range = M.get_cursor_range(win)
+    local range, cursor_range = M.get_cursor_ranges(win)
     lang_tree = M.language_for_range(lang_tree, range, config)
     if not lang_tree then
         return nil
@@ -62,9 +64,14 @@ M.get = function(bufnr, config)
         return nil
     end
 
-    local node = lang_tree:named_node_for_range(range, { bufnr = bufnr })
+    local node = lang_tree:named_node_for_range(cursor_range, { bufnr = bufnr })
     if not node then
         return nil
+    end
+
+    -- if the scope didn't change, return node immediately
+    if buffer_state.scope and buffer_state.scope:equal(node) then
+        return node
     end
 
     local excluded_node_types =


lukas-reineke:
> Importantly, it doesn't change the language_for_range calculation, which from what I can tell is the reason the other range was starting at column 0 of the current row.

This is not the only reason. This is also needed to filter out single line scopes. There were a lot of other edge cases with this logic, I should have written tests for this..

In principle, I'm fine with the change, but it needs logic to filter single line scopes. And _a lot_ of testing.

Danielkonge:
Just to make sure that I understand how you would want it to work.

My update gives behavior like this, since the scope is inside the function:

Screenshot 2023-12-27 at 16 14 30

The master branch gives behavior like this:
Screenshot 2023-12-27 at 16 15 03

Do you want an option to keep the old behavior? E.g., something like `one_line_scope` (maybe with a better name)? Because I would at least prefer to have an option actually underlines even for one line scopes. (I could underline the exact scope range in those cases too then.)

When I better understand the behavior you would want I can update this, and then I can add some tests after you have looked at it. (I would like to know if the code makes sense to you before adding the tests.)

lukas-reineke:
> Do you want an option to keep the old behavior?

The old behavior should be the default. I don't really want to add another option, maybe `show_exact_scope` can also enable single line scopes with just underline.

Danielkonge:
> The old behavior should be the default. I don't really want to add another option, maybe `show_exact_scope` can also enable single line scopes with just underline.

I have updated the code now so that the old behavior for one line scopes is the default and we only highlight one line scopes if `show_exact_scope` is true.

The code still fixes the highlighting issue described in #799 for `show_exact_scope` both true and false. 

Does this look good to you? Or do you want to change anything else?

sett-dozee:
When is this being merged to the main branch?

kuzicki (approved):

felixakiragreen:
I was about to report this as a bug, but I found the solution was already fixed and pull requested. Any chance we can get this merged?

PR #810, opened 2023-12-25
feat: repeat indent guides on wrapped lines with 'breakindent'

### LINKED ISSUE #577: Indent guides on wrapped lines
Word wrapping tend to cause the rendered guidelines to be broken up:

image

This used to be impossible to fix, but a recent PR made it so that you can now draw virtual text on wrapped lines with `virt_text_win_col`: 

- https://github.com/neovim/neovim/issues/23108#issuecomment-1565890709
- https://github.com/neovim/neovim/pull/23759.

It would be nice if this plugin can take advantage of this new feature.

Related issue: https://github.com/lukas-reineke/indent-blankline.nvim/issues/149

luukvbaal:
Opt-out with `ibl.config.indent.repeat_linebreak = false`.

Close #577

### DIFF
--- a/doc/indent_blankline.txt
+++ b/doc/indent_blankline.txt
@@ -328,6 +328,12 @@ config.indent                                              *ibl.config.indent*
 
                          Default: `1` ~
 
+   • {repeat_linebreak}  (boolean)
+                         Repeat indentation guide virtual text on wrapped lines
+                         if |'breakindent'| is set
+
+                         Default: `true` ~
+
  Example: ~
    >lua
    {
@@ -336,6 +342,7 @@ config.indent                                              *ibl.config.indent*
        highlight = { "Function", "Label" },
        smart_indent_cap = true,
        priority = 2,
+       repeat_linebreak = false,
    }
 <
 
--- a/lua/ibl/config.types.lua
+++ b/lua/ibl/config.types.lua
@@ -42,6 +42,8 @@
 ---@field smart_indent_cap boolean?
 --- Virtual text priority for the indentation guide
 ---@field priority number?
+--- Repeat indentation guide virtual text on wrapped lines if 'breakindent' is set
+---@field repeat_linebreak boolean?
 
 ---@class ibl.config.whitespace
 --- Highlight group, or list of highlight groups, that get applied to the whitespace
@@ -158,6 +160,8 @@
 ---@field smart_indent_cap boolean
 --- Virtual text priority for the indentation guide
 ---@field priority number
+--- Repeat indentation guide virtual text on wrapped lines if 'breakindent' is set
+---@field repeat_linebreak boolean
 
 ---@class ibl.config.full.whitespace: ibl.config.whitespace
 --- Highlight group, or list of highlight groups, that get applied to the whitespace
--- a/lua/ibl/autocmds.lua
+++ b/lua/ibl/autocmds.lua
@@ -29,7 +29,7 @@ M.setup = function()
     })
     vim.api.nvim_create_autocmd("OptionSet", {
         group = group,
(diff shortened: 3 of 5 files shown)


luukvbaal:
Depends on https://github.com/neovim/neovim/pull/26625

lukas-reineke:
Thanks, this is nice.

> may be replaced with a version check at some point I suppose.

Yeah, version check would be cleaner.

> One caveat here is that since extmarks are local to a buffer, and 'breakindent' is a window option, the indent lines may overlap buffer text when there are multiple windows with different 'breakindent' values displaying the same buffer... Doubt that is a real issue though.

This is fine, there are other option values with the same restriction already. I have this helper function to get the right window ID to use with `nvim_get_option_value`. Please use that as well.
https://github.com/lukas-reineke/indent-blankline.nvim/blob/0dca9284bce128e60da18693d92999968d6cb523/lua/ibl/utils.lua#L131-L142

And please add the option to the docs as well.

luukvbaal:
Hmm there is another caveat. Depending on the value of `'breakindentopt'`, we may still end up placing the indent guide virtual text over the actual buffer text. Not sure how to handle that TBH, here or in the nvim API.

doc/indent_blankline.txt
```diff
@@ -328,6 +328,12 @@ config.indent                                              *ibl.config.indent*
 
                          Default: `1` ~
 
+   • {repeat_linebreak}  (boolean)
+                         Repeat indentation guide virtual text on wrapped lines
+                         if 'breakindent' is set
```
lukas-reineke, line 333: This way it links to the `breakindent` docs
```suggestion
                         if |'breakindent'| is set
```

lukas-reineke (changes requested):
Just one small comment, rest looks good now

lua/ibl/config.types.lua
```diff
@@ -42,6 +42,8 @@
 ---@field smart_indent_cap boolean?
 --- Virtual text priority for the indentation guide
 ---@field priority number?
+--- Repeat indentation guide virtual text on wrapped lines if 'breakindent' is set
+---@field repeat_linebreak boolean
```
lukas-reineke, line 46: This has to be optional, that's also the reason the LSP tests fail.
```suggestion
---@field repeat_linebreak boolean?
```

lukas-reineke:
I will think about how to handle `breakindentopt` correctly. It's pretty niche, I'm okay with just not supporting this for now.

luukvbaal:
> I will think about how to handle `breakindentopt` correctly. It's pretty niche, I'm okay with just not supporting this for now.

That's my feeling as well. I'll address the comments/CI failure later.

lukas-reineke (approved):
thank you :rocket:

mehalter:
Should this be on by default? It seems like having the default options require the latest Neovim nightly is a bad idea even if the user can opt out of the new feature. It seems like it should be an opt in

luukvbaal:
Whether an option requires a certain Neovim version seems irrelevant to the option's default value. I think most people will want this option enabled, that's why I suggested opt out.

I do agree that just checking for `has("neovim-0.10")` is more disruptive than it has to be for  for people on a nightly version, but not the latest nightly. But they are using a nightly version, they can just update which I suppose was @lukas-reineke's view :)

mehalter:
Yeah I agree. I'll be honest, I just saw in the docs that the default is true and I didn't notice the check for 0.10. this is a completely sane default and relying on people using nightly to just update makes sense for me! People using nightly accept the chance for bugs and potentially infinite amounts of instability 😂

lukas-reineke:
The option will simply be ignored if you are on stable.
And if you are on nightly, it's your responsibility to be up to date.

PR #820, opened 2024-01-06
fix: remove starlark top level scope

Danielkonge:
From what I can tell this change should have been included in https://github.com/lukas-reineke/indent-blankline.nvim/commit/9301e434dd41154ffe5c3d5b8a5c9acd075ebeff since `module` seems to be the top level scope in starlark.


Danielkonge:
I noticed another problem, I will make a combined pull request with that too instead of this.

PR #821, opened 2024-01-06
fix: remove top level scopes

Danielkonge:
This removes the top level scopes for starlark and yaml, since they cause problems when scope highlighting.

Note: `chunk` is also a top level scope for Lua, but with the Lua parser `chunk` doesn't seem to cause any problems.


lukas-reineke (approved):
:rocket:

PR #822, opened 2024-01-06
feat: add git_config to scope_languages

Danielkonge:
This adds git_config to scope_languages. From what I can tell `section` is the only scope (the format is pretty simple).


lukas-reineke (approved):
:rocket:

PR #830, opened 2024-01-21
Temporary fix: eliminating the missing key 'virt_text_repeat_linebreak'

pshentsoff:
After the last update throws an exception on every cursor move. Disabling the missing key seems works fine - don't know what is it for, sorry. Maybe it'll be helpful.


lukas-reineke:
You are getting the error because you are using an outdated nightly version of Neovim.

PR #831, opened 2024-01-22
Update scope_languages.lua for Dart language

ralph-bergmann:
add some more types from tree-sitter-dart


lukas-reineke:
I don't know dart, but a lot of those don't sound like they would be scope :thinking:

ralph-bergmann:
I have tested them all, and they all make sense. Here are a few examples:

SCR-20240123-dxqx

lukas-reineke:
Scope has semantic meaning, please make sure you read `:help ibl.config.scope`.

There are places in your example that are not scope.

PR #871, opened 2024-05-17
fix: attempt to call field 'iter' (a nil value)

mcvnh:


lukas-reineke:
This is not broken.
The latest stable Neovim release is 0.10, `vim.tbl_flatten` is deprecated now.
IBL only supports the latest stable version and nightlight, 0.9 is no longer supported.
If you want to continue to use 0.9, please pin an older version of this plugin

PR #885, opened 2024-05-21
fix: highlights.lua setup builtin hl groups for all Ibl names

gavinbeatty:
IblScopeChar and IblChar were missing.

Without this change, it's possible to trigger the "No highlight group found" path for those names.
For example, I use the plugin junegunn/goyo.vim which when activated, pushes some theme/highlight changes to create a distraction-free writing environment.
When _exiting_ goyo mode, the missing name errors occur. I didn't inspect the code much, just noticed some of the builtin groups were not set in the relevant setup function, and adding them seemed to resolve my issue.

I don't know what the best initial values to setup would be, but I chose to copy from Normal.
My casual inference was the *Char was generic sounding enough to warrant copying "normal text".
My only visual inspection of the results is that it looks fine to my non-aesthete eyes.

It's possible that goyo is still doing something wrong here, but facially this seems like the correct thing to do.

### DIFF
--- a/lua/ibl/highlights.lua
+++ b/lua/ibl/highlights.lua
@@ -28,9 +28,12 @@ end
 local setup_builtin_hl_groups = function()
     local whitespace_hl = get "Whitespace"
     local line_nr_hl = get "LineNr"
+    local normal_hl = get "Normal"
     local ibl_indent_hl_name = "IblIndent"
     local ibl_whitespace_hl_name = "IblWhitespace"
     local ibl_scope_hl_name = "IblScope"
+    local ibl_scope_char_hl_name = "IblScopeChar"
+    local ibl_char_hl_name = "IblChar"
 
     if not_set(get(ibl_indent_hl_name)) then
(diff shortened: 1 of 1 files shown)


lukas-reineke:
`IblScopeChar` and `IblChar` are not missing, they are not correct highlight groups in the first place. They do not exist in this repo at all.  
Maybe the color scheme you are using has added them by mistake?

gavinbeatty:
Ah, I found those names by doing a recursive grep in my lazy directory, but didn't notice that IblChar and IblScopeChar come from NvChad plugins, not ibl itself.

- https://github.com/NvChad/ui/blob/91f5077e9268c8a8200c5f4b277328057ef3dd83/nvchad_types/all_hl_groups.lua#L291
- https://github.com/NvChad/base46/blob/7f9a7092402ae8951877f698f1c2291aa47c2903/lua/base46/integrations/blankline.lua

Fix belongs in base46. Sorry for the noise.

PR #906, opened 2024-07-05
doc: update README section for rainbow-delimeters integration

aliaksandr-trush:
We do not need to redefine highlight colors for rainbow-delimiters, but we can use already defined from the plugin. It will also integrate some themes out of the box.


lukas-reineke:
Thanks, but I think the more explicit version is better for the documentation.

PR #909, opened 2024-07-20
Add Perl to scope languages

pjcj:
I think these are the only two nodes we need.  I been using this for a little while and the only problem I've noticed is due to a treesitter mistake (https://github.com/tree-sitter-perl/tree-sitter-perl/issues/177).


lukas-reineke (approved):
thanks

pjcj:
Done

lukas-reineke (approved):

pjcj:
Thank you!

PR #919, opened 2024-09-02
chore: add doc/tags to .gitignore

liskin:
`:helptags ALL` creates this file and `git status` then keeps telling me about it. Other nvim plugins have this in their .gitignores, so add it here as well to silence it.


lukas-reineke (approved):
Thanks

Issue #926, opened 2024-10-04
Haskell (or "new" language) support for scope + surprising scope behavior


guibou:
### Problem

I've tried to add `node_type` for `haskell` without any result, until I manually edited `lua/ibl/scope_languages.lua` to add an empty reference for `haskell`.

Once it was done, I was able to have scope logic for a few node, but the support is partially bugged, some nodes do not works at all, some are partially displayed, more details in "Step to reproduce". I don't know if that's something `indent-blankline` can fix or if this is related to the treesitter support for haskell.
Nothing changed in the file

### Steps to reproduce

Let's assume the following haskell file:

```haskell
main = do
  print "hello"

  let
    x = 10
    y = 20

  res <- do
    print "hi"
    print "ho"
    pure 10
  
  pure ()

  where
    a = 1
    b = 2

blurk = do
  print "yoto"
```

Here is what I add to my `setup()`:

```
      include = {
          node_type = { haskell = { "do" } },
      },
```

The `do` node is located where `res <- do` is.

Without any other action, there is no change of highlight in my editor (note that scope works fine for rust, python, ...).

The following makes it work:

```
diff --git a/lua/ibl/scope_languages.lua b/lua/ibl/scope_languages.lua
index 496fcb3..c577666 100644
--- a/lua/ibl/scope_languages.lua
+++ b/lua/ibl/scope_languages.lua
@@ -1,6 +1,8 @@
 -- from nvim-treesitter/queries/{lang}/locals
 
 local M = {
+    haskell = {
+    },
     ada = {
         compilation = true,
         package_declaration = true,
```

So first, I suspect that the `include.node_type` configuration option only works for languages which are already listed in `scope_languages.lua`.

Now it works fine for `do` blocks:

![image](https://github.com/user-attachments/assets/8aa17d80-6b6d-4394-b657-b2cbedaa11e4)

And for the "top level":

![image](https://github.com/user-attachments/assets/a14e2bbf-7784-4363-9140-2c500a72ba1f)

I also wanted to scope the `where` and `let` clauses. I thought that `local_binds` node would be perfect for that:

![image](https://github.com/user-attachments/assets/563dc52b-bdf0-47ca-b042-7d2b64cd8877)

However, the result is not as expected (after adding `local_binds` in my `node_type` list):

On the first line of the `where` close, it matchs nothing:

![image](https://github.com/user-attachments/assets/2989d657-40a5-4e78-a945-613abd32aaaa)

[...]

PR #928, opened 2024-10-04
fix: scope logic with unknown language

guibou:
If the language is not listed in the languages "officially" supported by ibl, `scope[lang]` is `nil`.


guibou:
This is related to https://github.com/lukas-reineke/indent-blankline.nvim/issues/926 and recent https://github.com/lukas-reineke/indent-blankline.nvim/pull/927

lukas-reineke (approved):
Thanks

PR #934, opened 2024-10-27
feat: use faster version of `vim.validate` if available

MariaSolOs:
The form of `vim.validate` that takes a spec table as input was deprecated in https://github.com/neovim/neovim/pull/30855 and plugins should now use the faster version introduced in https://github.com/neovim/neovim/pull/28977.

This PR adds a new utility function to use the faster version if available. Also defined the existing `utils.validate` function to avoid confusion with the new utility function.

### DIFF
--- a/lua/ibl/utils.lua
+++ b/lua/ibl/utils.lua
@@ -12,7 +12,7 @@ end
 ---@param opt table
 ---@param input table
 ---@param path string
-M.validate = function(opt, input, path)
+M.validate_config = function(opt, input, path)
     vim.validate(opt)
     for key, _ in pairs(input) do
         if not opt[key] then
--- a/lua/ibl/utils.lua
+++ b/lua/ibl/utils.lua
@@ -9,11 +9,29 @@ M.get_whitespace = function(line)
     return string.match(line, "^%s+") or ""
 end
 
+--- Use the faster validate version if available.
+--- NOTE: We disable some Lua diagnostics here since lua_ls isn't smart enough to
+--- realize that we're using an overloaded function.
+---@param spec table<string, {[1]:any, [2]:function|string, [3]:string|true|nil}>
+M.validate = function(spec)
+    if vim.fn.has "nvim-0.11" == 1 then
+        for key, key_spec in pairs(spec) do
+            local message = type(key_spec[3]) == "string" and key_spec[3] or nil --[[@as string?]]
+            local optional = type(key_spec[3]) == "boolean" and key_spec[3] or nil --[[@as boolean?]]
+            ---@diagnostic disable-next-line:param-type-mismatch, redundant-parameter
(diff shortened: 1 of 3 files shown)


lua/ibl/utils.lua
```diff
@@ -9,11 +9,23 @@ M.get_whitespace = function(line)
     return string.match(line, "^%s+") or ""
 end
 
+--- Use the faster validate version if available.
+---@param spec table<string, {[1]:any, [2]:function|string, [3]:string?, [4]:true?}>
+M.validate = function(spec)
+    if vim.fn.has("nvim-0.11") == 1 then
```
MariaSolOs, line 15: I do confess that I'm not sure if this is enough since some early versions of 0.11 didn't yet have the new validation function, but I'm not too sure of how to more accurately check for this. Lmk @lukas-reineke if you have a better idea here.

MariaSolOs: I also think this is what's causing the failing CI checks...

MariaSolOs:
I think that the failing CI checks are because they're not using the updated `vim.validate` overloaded type.

lua/ibl/utils.lua
```diff
@@ -9,11 +9,25 @@ M.get_whitespace = function(line)
     return string.match(line, "^%s+") or ""
 end
 
+--- Use the faster validate version if available.
+---@param spec table<string, {[1]:any, [2]:function|string, [3]:string|true|nil}>
+M.validate = function(spec)
+    if vim.fn.has "nvim-0.11" == 1 then
+        for key, key_spec in pairs(spec) do
+            local message = type(key_spec[3]) == "string" and key_spec[3] or nil
+            local optional = type(key_spec[3]) == "boolean" and key_spec[3] or nil
```
lukas-reineke, line 18: lua-ls isn't smart enough to do this type narrowing automatically, `message` and `optional` are still `string|true|nil` and CI complains. You can just cast it here.

```suggestion
            local message = type(key_spec[3]) == "string" and key_spec[3] or nil --[[ @as string? ]]
            local optional = type(key_spec[3]) == "boolean" and key_spec[3] or nil --[[ @as boolean? ]]
```

lua/ibl/utils.lua
```diff
@@ -9,11 +9,25 @@ M.get_whitespace = function(line)
...
+            vim.validate(key, key_spec[1], key_spec[2], optional, message)
+        end
+    else
+        vim.validate(spec)
```
lukas-reineke, line 22: This should make CI happy
```suggestion
        ---@diagnostic disable-next-line
        vim.validate(spec)
```

MariaSolOs:
> just some small notes to fix CI.

Thanks for the hints! :)

I also want to say that I'm very impressed by your CI. I've contributed to quite a few plugins and this is a very complete testing suite!

lukas-reineke (approved):

PR #942, opened 2024-11-11
refactor: remove redundant conditional

wincent:
Removes one of the nested conditionals:

```lua
if config.exclude then
    if config.exclude then
```

### DIFF
--- a/lua/ibl/config.lua
+++ b/lua/ibl/config.lua
@@ -231,7 +231,6 @@ local validate_config = function(config)
         end
     end
 
-    
     if config.exclude then
         utils.validate_config({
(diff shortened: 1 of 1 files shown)


lukas-reineke:
Thank you, can you format the file with `stylua`? Then CI should be happy

wincent:
> Thank you, can you format the file with `stylua`? Then CI should be happy

Done. That's what I get for editing the file directly in GitHub's web UI. 😬

lukas-reineke (approved):

lukas-reineke:
Tnank you :heart:

PR #943, opened 2024-11-11
fix: correct copy-pasta in config.indent validation

wincent:
Looks like we're passing in `char` when trying to validate `tab_char`.


lukas-reineke (approved):
thanks

